PR #5, opened 2015-01-03
A slightly better and more straightforward way to workaround the bug in the po4a

Fat-Zer:
Here is a slightly better and more straightforward way to workaround the bug in the po4a: to add an alt text for the images....


ciampix:
Thanks for the po4a patch too. Hope it will be fixed soon.

PR #8, opened 2015-02-04
Transcribe Pcbnew EN doc, first 2.5 chapters, basic markup, very minor error corrections

johnbeard:
Some things, such as the 1/10,000 inch resolution are out of date totally. I have not changed these as I don't want to do the format shift and content changes in the same commit.

I have used separate .adoc files as "chapters" to avoid a a monster file (and to allow modification area to be seen at a glance by filename), but GitHub doesn't seem to handle it, though Asciidoctor.js and the compiler seem fine.

Note: some of the images are saved as .gif in the .odt files. I have converted to .png.

Also it has a pcbnew front cover, not KiCad that I took from the pt translation (I think). Maybe that should be the plain KiCad one for now?


ciampix:
In the mean time I pull your request. For the image I have to investigate, I do not have it here the odt at hand...

PR #13, opened 2015-02-26
Transcribe pcbnew chapter 6-8 (en)

johnbeard:


ciampix:
Many many many thanks!

PR #21, opened 2015-03-28
Finishing Interactive Router manual translation

keruseykaryu:
BTW: There is something wrong with the unordered lists in this file. Only English edition is correct. Both Italian and Polish edition does not render these lists.


ciampix:
There is a small bug in po4a that can be circunvented adding some spacing in unordered lists. Would you point me the wrong rendered lists so I can fix them? Thanks!

ciampix:
Nevermind, done!

PR #22, opened 2015-03-30
Added doc for new annotation option

cpavlina:
Added the documentation for "Reset, but do not swap any annotated multi-unit parts", which I added recently. Also replaced the "annotate" toolbar icon with a new one, as the icon has changed.


ciampix:
Thanks!

PR #24, opened 2015-04-05
Create initial translations for Eeschema manual.

keruseykaryu:
Add Polish localized images to the other reference manuals.
INFO: Some of Pcbnew localized images still aren't properly renamed to match it with their English equivalents.


ciampix:
Thanks!

PR #32, opened 2015-04-24
Eeschema manual has been fully (re)translated

keruseykaryu:
Eeschema manual has been fully (re)translated.

Section 9.5 (at Eeschema_create_a_netlist.adoc) is not reflected in the .po file. I don't know why. Probably by the "&raquo;" character in the end of the header.


ciampix:
I've removed such chars. Please check since there are more than one fuzzy signalled and I have _not_ modified anything else from our last "troublesome" conflict. I have decided that I won't improve the original file before you have finished working on your translation to avoid nasty conflicts.

keruseykaryu:
> Please check since there are more than one fuzzy signalled

I can not find any fuzzy translations in my already translated manuals. 100% has been translated.
I didn't work alike on any other translation now, so IMO there is no chance to lose anything.

ciampix:
Having modified the source file you _should_ have at least one fuzzy string. Try forcing an update doing ./make.sh po-all. Sorry I am not a programmer so I doubt that my Makefile could not be improved.

ciampix:
I am working on Pl_Editor and there there is no pl translations, just polish screenshots. I try to rename the unmeaningful english image names into file with the same name that you already had used, renaming just when I have to do it (when the name is incorrect or wrong) just to make your (and other translators) life easier.

keruseykaryu:
I've updated Eeschema translations.

PR #34, opened 2015-04-25
Renamed images to meagniful names for KiCad manual

keruseykaryu:
BTW: I found that you left out the table at line 224 in Italian translation.


ciampix:
BTW today is a national holiday in Italy. The end of the II world war and liberation from nazi-fascism. The start of the liberation was due to the Polish army sacrifice in the infamous "Monte Cassino Battle", see https://pl.wikipedia.org/wiki/Bitwa_o_Monte_Cassino

ciampix:
Great work, thanks!

PR #35, opened 2015-04-26
GerbView manual has been translated

keruseykaryu:
The GerbView manual has been translated, but I couldn't find what is the problem during PDF creation.

error

I was trying to retranslate and rewrite this string in some different ways, but this isn't help. Even if this particular translation is removed (empty string), the issue still exist and point to the source string.


ciampix:
It is probably due to some "strange" char in a section title/header ... augmenting verbosity and seeing in temp files usually helps... I will see about it in the next hours...

ciampix:
pdflatex it is far from perfect. Hope that asciidoctor direct pdf conversion will not take too much to finish...

keruseykaryu:
I think I fix this in https://github.com/ciampix/kicad-doc/pull/36
Now everything should back to normal.

PR #36, opened 2015-04-26
Remove double quotes which may cause strange issues

keruseykaryu:
Currently problem occurred only in case of Polish translation.


ciampix:
Already converted "" into ``''. Seems that pdflatex do not like simple double quotes...

ciampix:
PS: thanks to Nick O.

PR #41, opened 2015-05-16
Big 'Getting started' update request

danger89:
Add a new workflow diagram to the Getting started documentation, including the Dia source file (*.dia). Plus updated several outdated screenshots with up-to-date ones, and updated the text accordingly.

Also updated the text and several links for how-to install Kicad.


keruseykaryu:
You use wrong acronym for Page Layout Editor. The correct is `Pl_Editor` with `L` letter, not `I` letter.

ciampix:
All ok, thanks!

nickoe:
@keruseykaryu I fixed that in 34667aac26576f9cac2b8cfe980dfafb04d9889b

nickoe:
Also, @danger89 and @ciampix, why was all the svg in src/Getting_Started_in_KiCad/images/flowchart_src/bitmap_sources added? Does it make sense to add them all, and in that folder?

keruseykaryu:
Besides what _Nick_ asked, I would like to ask why whole icon set was added to the Getting_Started part?

I think the include of whole icon set being good when we create a one separate directory accessible by any manual. This way we don't need to manage icons for particular manuals.

danger89:
@nickoe I will see how the flowdiagram looks like on A4. If needed I will update it.

@keruseykaryu @nickoe Good question again: I just saw that there were using png files. So I converted the latest svg icons to PNG's. And replaced the whole icon set to be up-to-date. A better approach in the (near) future would be: a single place of icons in the whole kicad-doc repo (root level?).

You can remove the bitmap sources if you want. You know know I used the svg files in the Dia figure.

ciampix:
As I wrote on irc unfortunately placing the images outside the document root dir break the epub generation. As Nickoe pointed out, using a main root image dir could still work if we manage to copy it inside the document root dir during epub compilation or we have to do away with epub generation.

ciampix:
I'll try to do some experiments to test if this is feasible.

danger89:
Ok thx..

PR #47, opened 2015-05-19
CMake Build system

BrianSidebotham:
There's the chapter globbing left to go, but this change adds a more-or-less complete CMake build system to the documentation. :)


nickoe:
I will merge this keeping in mind that it does not really include the includes. I have not investiged why yet.

But shouldn't a2x just do that per default?

BrianSidebotham:
That's right, this needs sorting out. The include files need to be translated and then get renamed, thus the includes need to be modified in order to fix the include problem.

I should be able to get round to it soon so that the complete build works.

I also wanted to have monolithic and individual chapter targets. For PDF a monolithic version is best and for HTML some people will prefer individual chapters.

PR #59, opened 2015-05-25
* Fixes #58 - provides SINGLE_LANGUAGE option when configuring

### LINKED ISSUE #58: CMake Builds always packages all languages
With the CMake build system, when packaging via `make package` all languages are included; There's no possibility to only build a package for a single language.

BrianSidebotham:
Adds a single language option. Safely builds a language, even partially translated languages such as `fr`, only what is translated is included in the build.

To use this commit, configure a single language build with:

    cmake -DSINGLE_LANGUAGE=en

Available languages so far are `en|fr|it|pl`

The package name is transformed when using **SINGLE_LANGUAGE** so that the language is included in the package tarball.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,10 @@ set( PDF_GENERATOR "DBLATEX" CACHE STRING "Specify which PDF generator to use: D
 
 # Provide an option for a single language build. If nothing else, this will at
 # least help developers to speed up a build they're working on
-set( SINGLE_LANGUAGE "" CACHE STRING "Specify a single language to build: en|fr|it|pl" )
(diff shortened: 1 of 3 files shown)


BrianSidebotham:
I've fixed #60 which meant you couldn't build `en` as the **SINGLE_LANGUAGE** option. I've also added a new configuration option **BUILD_FORMATS** which is a list of build formats to build. Use it to specify a list of build formats (as we only have two formats, this option is simply for selecting only one of them.)

`-DBUILD_FORMATS=html` or `-DBUILD_FORMATS=pdf`

BrianSidebotham:
If the ci server fails again after the new commits, please paste the output somewhere. I don't think I have credentials for the ci.

BrianSidebotham:
Oh, okay I see it was just a Java face plant. I'll wait for the next job to complete.

PR #74, opened 2015-06-09
Add documentation for component library rescue

cpavlina:


nickoe:
Thank you for your contribution.

I don't quite know why, but it seem Jekins is not that happy, but it looks good to me and builds locally, so I will merge it anyway.

ciampix:
Fixed, thanks!

PR #76, opened 2015-06-09
Tidy up some of Eeschema documentation

cpavlina:
The language in the Eeschema documentation is pretty awkward. I've started to go through and rewrite the text with a more English-native and technical documentation feel. Let me know if you don't like the changes, otherwise I'll proceed with the rest of it. I'll update old screenshots as I go, and probably redo the documentation for things that are out of date (like the BOM export tool).


keruseykaryu:
You can also improve the style a bit, so the text does not look like written from the first person perspective: "You can click" -> "By the clicking", etc.

cpavlina:
"By clicking", you mean. Yes, that also works, but often leads to more long-winded explanations. I've found people generally do better reading documentation written in a slightly less formal manner.

(However, I can change this if desired.)

ciampix:
Manually commited thanks!

PR #85, opened 2015-06-21
Updated IDF documentation

cbernardo:
I have made some updates (notes), fixed a few errors and added comments on currently known bugs.

I think the formatting of some images (for example the SolidWorks rendering of IDF) could be improved in the html document (size the image to match text width) but I'm struggling to find information on how to do this. (Looking for an equivalent to LaTeX's  "width=0.8\textwidth".) Some images in the PDF doc can also be improved by using their natural size rather than stretching to fit the page width.


nickoe:
I agree that some image handling probably need to be done, but this is for all of the docs, so lets call that another work package for someone. I will merge this.

PR #86, opened 2015-06-21
Partial rewrite of the section on DXF import in Pcbnew.

MarcoHess:
The notes on missing segments in the Ellipses shapes no longer applies as this is now fixed in Pcbnew.


MarcoHess:
Not clear as to what the failure is, but here is some further improved wording.

MarcoHess:
I don't understand what this error is. It seems to occur in the french 
translation?

On 21-Jun-15 14:29, kicadjenkins wrote:
>
> Refer to this link for build results (access rights to CI server needed):
> http://ci.kicad-pcb.org/job/kicad-doc-testing-pull-request/58/
> Test FAILed.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/ciampix/kicad-doc/pull/86#issuecomment-113864682>.
>

-- 
Marco Hess
Through IP Pty. Ltd. - AUSTRALIA
www.through-ip.com  | marco.hess@through-ip.com
p: +61 407 78 55 66 | f: +61 8 8121 6191

nickoe:
It seems to be this error that triggers the fail:
```
00:08:05.843 asciidoc: ERROR: Pcbnew_create_and_modify_board-fr.adoc: line 87: undefined filter attribute in command: source-highlight -f {basebackend-xhtml11?xhtml}{basebackend-html4?html} -s {language} {src_numbered?--line-number} {src_tab?--tab={src_tab}}
```

Do you know anything about this Ciampix?

nickoe:
I suspect it to be the pings around for example `'advanced'`. Does this build fine locally for you?

MarcoHess:
I did not build it locally. Pings around a word are not allowed? What do I need to use 'emphasize' a word?

MarcoHess:
Ok, I read up a bit more on the asciidoc format and changed the emphasis used a bit. Also, some of these emphasis items were used at the end of a sentcence and I included a space in those case.

cpavlina:
Are you testing this locally by running make.sh? You should be able to pick up this stuff before committing - the cmake build still doesn't catch everything, though, because it doesn't build all formats.

MarcoHess:
No I am not. I am developing on Windows and so far have not been able to 
successfully build ANY documentation.

I find it particularly frustrating that there is no proper error message 
to indicate what is wrong.

The error seems to be generated during the french language build ?!?!?! 
but does not say what or where (that I have been able to find at least). 
The error also seems to be some JAVA crash complete with back trace, so 
what is the deal with that?

With all that, how do I figure out what is the problem then with the 
English text? Is there actually a problem? I really can't see what the 
problem could be in just some paragraphs of text with a just a few 
emphasised or bold words.

Marco

Caerbannog:
By trial and error I found that the problem is the two spaces on this line:
https://github.com/MarcoHess/kicad-doc/blame/3205b6a8f1066b0050da3f6828c53a85ebfc1482/src/Pcbnew/Pcbnew_create_and_modify_board.adoc#L61

They cause po4a to create a translation block with two concatenated lines as such:
```
Click OK Notes on importing DXF files with advanced features like Ellipsis
```

Then the underline for the title (`^^^^^^...`) doesn't match, and I have not investigated more.

The solution is to remove the spaces on line 61.

Caerbannog:
The French translation isn't especially responsible, it's just that it gets built first I think.

MarcoHess:
I cleaned up all trailing spaces in that file. Hopefully this fixes it now.
Also reworded slightly again :-)

nickoe:
Hmm, it seems to be fixing it, so I will merge. I would not have expected the translations to get confused by trailing whitespace, I would just expect it to interpret it as an extra space.

But thank your for your contributions and fixes.

PR #94, opened 2015-06-22
Remove spaces that cause the image code to be displayed instead of the picture

Caerbannog:


nickoe:
Hmm, funky behaivour, but thank you. I will merge it.

Issue #96, opened 2015-06-25
Icons in eeschema 11.3.1. Main Toolbar is outdated


nickoe:
Icons in eeschema 11.3.1. Main Toolbar is outdated.

Anyone, fell free to fix.

nickoe:
The commit merged above fixes the original issue, but FYI there might still be places where this is still an issue, but I will mark this a closed.

PR #98, opened 2015-06-28
Enforce the distinction between icons and generic images

Caerbannog:
Icons are not consistent in the documentation : sizes vary, some have backgrounds depending on the OS, some have underlines if the menubar was not cropped after a screenshot.
Another problem is keeping icons up-to-date with KiCad evolution (see issue #96).

Proposed changes:
- Icon names should be the same as in KiCad, this will help to stay in sync:
https://github.com/KiCad/kicad-source-mirror/tree/master/bitmaps_png/cpp_26

- Icons should go in the "images/icons" directory instead of being spread over images, images/en, and images/icons.

- Icons should be copied from the KiCad build tree when possible. This will ensure consistency.
    To generate the PNG when building KiCad, you have to set `MAINTAIN_PNGS:BOOL=ON` in `CMakeCache.txt`.

This pull request is to check that such changes are welcome. Lots of icons are not yet updated.


ciampix:
Very well done, thanks!

PR #100, opened 2015-07-05
Add japanese translation

yoneken:
Dear @ciampix ,
Many thanks for your leadership you provided in KiCad translation.


yoneken:
Ooops..
The error is come from the japanese font missing..
OK, I will tackle against the problem.

nickoe:
@yoneken Hi, thank you for your work. I was trying to look into why this fails, and it seems to fail with the way the old makefiles does it, called with `./make.sh`. Can you try to run that in the CvPcb folder and see if it will fail for you too?

It looks like there are some latex depends that we are missing or maybe we need to add something to the preamble.

yoneken:
Thank you @nickoe , I fixed the issue.
However, to build japanese pdf, the system needs to install fonts-vlgothic, which is a font of japanese.

nickoe:
@yoneken it does indeed seem to build, but there is still a problem with the resulting pdf. It will show every char as a rectangle. And in the cmake generated pdf I get stuff like `CvPcb &#x306e;&#x7279;&#x5fb4;`  ... yes the unicodes printed. It looks like the encoding that could possibly exist from an xml structure. (I think a2x will output that, maybe docbook).

yoneken:
@nickoe It seems to no problem!
Voila! In my environment, I can read the correct PDF.
http://kicad.jp/translate/screenshot.png

I guess, if you install some japanese font (fonts-vlgothic etc..), you can see the same result.

nickoe:
I did install the font, and I can use i it other programs like inkscape and the html output looks fine. I think it might be some issue with the PDF not embedding the font somehow.

nickoe:
The build server do have the fedora vlgothic-fonts installed, FWIW. :/

nickoe:
Ok, it sort of seems ok, but not completely. The cmake stuff is still strange printing unicodes, whilst the old make system works. I guess we just miss some options to a2x in the cmake stuff.

Cmake: http://ci.kicad-pcb.org/job/kicad-doc-testing-pull-request/81/artifact/build-cmake/src/CvPcb/ja/CvPcb.pdf
Old: http://ci.kicad-pcb.org/job/kicad-doc-testing-pull-request/81/artifact/src/CvPcb/CvPcb-pl.pdf

yoneken:
OK. I confirmed the issue.
I cotinue to fix it.
Thank you for your kindness!

yoneken:
I fixed the issue of cmake build.

However, I realized that there is another issue.
Helps of Eeschema, GerbView, and Pcbnew include chapters from main adoc.
The link names in main adocs are pointed to original (english) adocs.

yoneken:
Is there any issue for merging this PR except #101 ?
Let me bring forward to suggest to pull the japanese translation.

BrianSidebotham:
After merging #102 you'll indeed get a translated copy, but we must pick a font for Japanese support. This will mean extending the CMake stuff a bit so that the font can be overridden I guess. Currently, I get reasonable Japanese output by installing the `apt-get install fonts-takao-mincho fonts-takao` packages and then adding `\setmainfont{TakaoExGothic}` in `src/pdf-cover-dblatex.sty` which we use for all PDF manuals.

It's not full though, there's still missing characters.

We need to extend CMake and add font dependencies because CJK manuals should compile on the main KiCad build server which won't be a CJK-specific version. Otherwise we just end up with a trashed manual.

BrianSidebotham:
Here's what it looks like with the fonts mentioned:

CvPcb in Japanese

nickoe:
What about that vlgothic font? @yoneken did port the xetex option to dblatex in 89a47b3fa631f3540f089f89fbcf5a9bc7ecfae1. It uses style/dblatex-pdf-ja.xsl which specifies the VL Gothic font, which I think looks good. (I don't know how japanese should really look like)

nickoe:
test this please

BrianSidebotham:
I can have a further look. If we bring in xetex (instead of pdflatex) for the Japanese language, we may as well use that backend for all languages. It doesn't need to be different for Japanese. I'll do some testing with the xetex backend for other languages too. We would need to add xetex to the dependency list and make sure we can test for it being present in the CMake build system.

I notice that what's missing in the font is a different formatting, most likely bold or something, hence characters that cannot be rendered is actually section titles. Some more `.sty` foo may help to fix that.

nickoe:
I think using the font mentioned previously will fix that. The xsk file specifies it for the frontmatter.

nickoe:
```
00:05:15.524 a2x -f epub -a lang=ja -a docinfo Eeschema-ja.adoc
00:05:27.905 a2x: ERROR: missing resource: /home/kicad/workspace/kicad-doc-testing-pull-request/src/Eeschema/images/ja/10000000000000DA000000C6E73B58D7.png
00:05:27.915 ../Makefile:187: recipe for target 'Eeschema-ja.epub' failed
```

yoneken:
Oops!
I pushed the missing image now.

yoneken:
I'm sorry.. This request remains many missing links in Pcbnew..
Tomorrow, I wii fix it.
Thanks.

nickoe:
It seems to be working now! Shall I merge it or do you want some other upates with this PR?

Btw @BrianSidebotham it seems that we include the adocs from the other language in the subfolder of a language. http://ci.kicad-pcb.org/job/kicad-doc-testing-pull-request/90/artifact/build-cmake/src/Pcbnew/ja/    Not critical, just not so nice.

yoneken:
I got it!
Thank you for your help!

Could you merge it?

nickoe:
Done. Thank your very much for your effort. Feel free to create new pull requests for followups.

yoneken:
Thank you!

In previous my comment,
> I got it!
Not "I", but "We".

tkamada:
There might have been a confusion on Japanese support of DocBook.
The setting of fonts by @yoneken was reflected to my advice.
I don't know so much on both DocBook and LaTeX mechanism behind it, then I told @yoneken only for using xelatex, not for pdflatex.

Though I couldn't catch up the current status of the result PDF on Japanese translation, does this issue resolved?

The key point of creating PDF from DocBook in current environment in Japanese stuff is:
* using xelatex, not pdflatex
* setting CJK fonts using CJK package included in the popular TeX distributions such as TeXLive and MikTeX.  Please never set global font into Japanese one because it makes ugly look to in the resulting PDF due to  different designs between original English glyph and Japanese glyph corresponding for English letters.

PR #103, opened 2015-07-08
fix missing '[]' of image: and double byte double-quote in pcbnew.adoc

kinichiro:
fix missing '[]' of image: and replace double byte double-quote with single one.


nickoe:
Thank you.

PR #110, opened 2015-07-10
reformat GerbView section 2 and 3.

kinichiro:
it is hard to read section 2 and 3 of GerbView
because of its less formatting.
Would you like formatting like this PR ?
I think this is better, but we need to translate .po of GerbView again.
Changes are a lot.

PR #113, opened 2015-07-12
Remove the old gerber explnations

nickoe:
This pull request is to supercede pull request #110


kinichiro:
Line 219, 
It is possible to **load** gerber and ...
is better.

kinichiro:
I think it is OK.
Thanks.

PR #116, opened 2015-07-15
Fix #107

### LINKED ISSUE #107: Build is broken
I think the Japanese merge ( #100 ) was too quick, now the build fails for me with various errors to do with xelatex and fonts. Please update the cmake-readme.adoc documentation to detail any added dependencies and also please fix the build.

The add_dblatex_option() works globally, so now all languages are being built with the xetex backend and with the japanese style sheet.

The japanese style xsl should go in the already-existing /xsl folder or else move (and re-reference) the kicad.xsl (used by FOP) file from the /xsl folder into the new src/style folder.

As a side, to support building with FOP we may need more style sheets, but we can probably deal with that at a later date.

yoneken:
Dear @BrianSidebotham , @nickoe ,
I added the required information about building japanese pdfs to fix #107 .

I would appriciate if you have a broad-mindedness for minority languages.


BrianSidebotham:
Thanks for this!! It's a start and that's really all I wanted - to not feel alone in making the CMake build system. Even just making notes like this saves people (including me!) a lot of grief. I have so little time available to maintain it at the moment.

I suggest we pull this in, then I can do some pull requests while I'm doing the version CMake stuff. It'll be good to remove the `-ja` part of the stylesheet and put the dblatex option call into the main `CMakeLists,txt` because it is applied to all languages anyway.

CMake should check for the font and barf if it cannot be found because it's a dependency for all languages. Windows currently has to use FOP and that doesn't barf on building the `ja` docs - it just ends up with garbage because the font glyphs are missing. On Windows we can use "Arial Unicode MS" which is a FULL Unicode font, but that requires more xml configs and/or xsl for FOP.

I also thought of removing the `SINGLE_LANGUAGE` option because it makes more sense to have a `LANGUAGES` option where you can list the languages to build.

Lastly, I cannot tell you how much your Japanese translation is appreciated. Thank-you so much for your effort, it's awesome! We should all feel proud that we now have a Japanese version of the documentation.

nickoe:
@BrianSidebotham You are not alone. I am kept busy reviewing the other contributors works and testing it. If you want someone else actually implement some cmake tasks, then I suggest creating an issue on it such that we can have an "easy" task to follow and implement. Then everyone with the time can work on it and it can be merged.

nickoe:
But I will merge this pull request.

Also, @yoneken I am not sure what you mean by the "broad-mindedness for minority languages". I personally don't see a hell lot of advantage of translations to my mother tongue, but that does not mean that your particular language will not matter. I love to see the docs used, and therefore I try to support every initiative for exposing the content.

But we might want to hold back with the translations and make sure that the english is up to date, such that translators does not need to update a section over and over again.

PR #119, opened 2015-07-20
Add lang-ja.conf and lang-pl.conf

Caerbannog:
Fix asciidoc "missing language conf file" errors.
Copy these files in $USER/.asciidoc or follow these instructions:
http://www.methods.co.nz/asciidoc/chunked/ch26.html#X27

Polish source: Kerusey Karyu
https://github.com/asciidoc/asciidoc/pull/65
Japanese source: WATANABE Yuki
https://github.com/asciidoc/asciidoc/blob/master/lang-ja.conf

I believe we should ship those files until they are integrated in most distribution packages.


nickoe:
Since it the link you refer says:  "The directory containing the AsciiDoc source file."

Maybe it could make sense to add it to the cmake configuraiton?

Caerbannog:
I added the language config files in the cmake build as you suggested.

This pull requests also includes a reorganization of the READMEs and small changes to make it more appealing to new contributors.

nickoe:
FYI, I will test this before merging. I expect to have time looking at it tonight.

nickoe:
Thank you very much. Now merged. There were minor bits here and there I might change, but this is definitely an improvement.

PR #125, opened 2015-07-24
Prepare webdocindex.adoc to be served at docs.kicad-pcb.org

Caerbannog:
Assorted commits: Compile GUI_Translation_HOWTO and webdocindex, update webdocindex, remove unused texlive dependencies

The aim is to replace http://www.kicad-pcb.org/display/KICAD/KiCad+Documentation with an index of KiCad documentation generated from asciidoc. Then to have it built automatically and served from:
http://docs.kicad-pcb.org/

We can eventually add URL prefixes like /stable, /v1, ...

This is a work in progress, with feedback very much welcome.


nickoe:
I notice that the webdocindex is put in a folder name "en" in the root of the src dir in the cmake build dir.

Should we change that?

I also see that you added some language flags, but I can't see where those are used or supposed to be used. Can you explain that?

Caerbannog:
Thanks for pointing both issues.

The "build/en" dir isn't a design choice. What would you think of moving it to root/index.html ?
Whatever would integrate easier with how docs.kicad-pcb.com is generated.

The flags were a bad idea, I will remove them.

nickoe:
I was searching for ways to include files in html5, I found something like:
  <link rel="import" href="/path/to/imports/stuff.html">
But it seems not to work out.  Alternatively we could possibly pust a php file, which will works on the server and be good enough for now.

Caerbannog:
Yes, and there are other solutions (add an install rule, create symbolic links, URL rewriting), they only depend on how the server is configured. Who manages it?

I think having a beautiful home for the docs on the web will really help user adoption.

nickoe:
I think I will merge as is, and then make the main build job copy .../webdocindex.html to .../index.html.

I in the future I expect it to be more integrated with the new webiste of kicad-pcb.org. So I think this solution should be good enough for now.

PR #127, opened 2015-07-24
Compile epub with cmake

Caerbannog:
Compilation of epub with CMake the same as for PDF. I reused the docinfo cover page from the legacy Makefile.
Hopefully there is not much more to do before removing the legacy Makefile and related make.sh

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ set( PDF_GENERATOR "DBLATEX" CACHE STRING "Specify which PDF generator to use: D
 set( SINGLE_LANGUAGE "" CACHE STRING "Specify a single language to build: en|fr|it|ja|nl|pl" )
 
 # Provide an option to decide which documentation formats to build.
-set( BUILD_FORMATS "html;pdf" CACHE LIST "Specify which output formats to build html/pdf" )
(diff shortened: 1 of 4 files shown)


kinichiro:
Japanese epub outputs seems not good.
All the Japanese characters are garbled.
html and pdf are OK.

nickoe:
Thank you, it seems to work ok, so I will merge it. Keep the fixes comming.

nickoe:
@kinichiro In what way? I opened the one in the artifacts for the builds, and it seems ok to me. Not that I can read japanese.

http://ci.kicad-pcb.org/job/kicad-doc-testing-pull-request/118/artifact/build-cmake/src/Pcbnew/ja/Pcbnew.epub

kinichiro:
It looks like this on my PC.
pcbnew_ja_epub_pdf

kinichiro:
when generating epub with `./make.sh Pcbnew-ja.epub` under src/Pcbnew/ directory,
I could get fine epub file, like this.
pcbnew_ja_epub_ok
Then, I think cmake Japanese epub build has some issue.

nickoe:
Ok, so we have some more font issues here.

nickoe:
I guess this can help us: http://www.methods.co.nz/asciidoc/faq.html#X5

PR #128, opened 2015-07-24
Add language option for epub

Caerbannog:
Try to fix Japanese epub from PR #127 
@kinichiro is there a progress?


yoneken:
This commit seems to fix the issue of japanese epub.

kinichiro:
@caerbannog
sorry, I slept over a little bit.
Thanks it is alright as @yoneken said.

nickoe:
I would have thought we needed to use the `--resource` option. But since it works, I will just merge it.

PR #130, opened 2015-07-25
Convert GerbView to 'article', it has only one chapter

Caerbannog:
Since removal of gerber specs, there is only one Level 1 title.


ciampix:
Thanks!

Issue #63, opened 2015-05-27
CMake package version is always 1.0.0


BrianSidebotham:
The CMake build system package target `make package` always uses the same version number - 1.0.0

1. How should the numbering system go?
2. Should it have a version at all?
3. Will it be aligned with the KiCad "stable" product?

ciampix:
IMHO I think that it should be aligned with the KiCad version it refers to (like GIMP manual for instance).

BrianSidebotham:
I also think keeping aligned is a good idea. I wonder if we should mark the version as the git branch. That way, a build from the master branch will get you `kicad-doc-master.tar.gz`, etc. and then when a stable release is almost ready with a known triplet, we can create a new branch such as `4.0.0` or whatever. That branch will end up with a package `kicad-doc-4.0.0.tar.gz`, etc.

This is probably the best way of keeping aligned and is also a good way of being able to work on a stable version's documentation.

The other option is to use tags. I'm sure there's a way of git getting the tag, coupled with the number of commits since the tag so we could end up with normal builds being the last stable tag, plus a number indicating the number of commits since that tag, such as `kicad-doc-4.0.0-15.tar.gz`, etc. That may also be good. I prefer branches to tags, but it's really the preference of the translators that matters most here I guess.

nickoe:
Instantly I am more inclined to use tags, and you should be able to look up the tag from some commit.  But when I think about it, I guess it is better to use branches, because the docs could possibly lag the release, and then updates specifc to that version could added just to that branch and alternatively also applied on the master.

mangelajo:
Thinking of autobuilds, it's easier to work with branches.

We could build testing to http://docs.kicad-pcb.org/testing/....
and stables to http://docs.kicad-pcb.org/stable-xx.xx.xx/.......

Please think of autobuilding and having everything available online in an automated fashion, no human interaction but pushing to branches... :)

I'm all for tags if that can be archived also with tags, then I don't care ;)

danger89:
I agree with Brian. The Github tagging is the most logical approach in this case. It makes more sense to create versions, then create multiple branches.
@nickoe  Branches are used to create separate copies and develop in a seperate branch, later you can merge it to the master branch. In this case you always want to have a history, so please choose for tags.

The tags/versions of the document corresponds to a version of KiCad.

danger89:
Maybe take a look at Linux as an example:
https://github.com/torvalds/linux

"Better well stolen than badly invented." They say here.

[...]

PR #134, opened 2015-07-26
Clean images in GettingStartedInKiCad

Caerbannog:
This splits icons/ in a dedicated directory. They are freshly generated, and will soon be shared between documents.

All auto generated image names are now gone.


nickoe:
I thought the idea was to have a global directory for the icons.

Caerbannog:
Yes, I just haven't written a CMake solution for the ePub problem of relative images. This commit will help though, splitting icons from images will be a necessary step.

nickoe:
I guess for a proper epub file we should embed stuff.

Caerbannog:
I think images are always embedded in what they call an OEBPS package, and the root of this package is the .adoc file directory. During the build anything higher is discarded by a2x as being outside of the hierarchy.

Caerbannog:
@nickoe In this rework I moved the icons to a global directory. Summary:
- Remove previous icons from /src/GSIK/images.
- Import recent PNGs to /icons (outside of /src), with a README explaining the generation.
- CMake now duplicates this directory for each translation of each document.
  I could not find a workaround because EPUB genration with `a2x` does not support relative paths with "../",
  and HTML generation with `asciidoc` will interpret paths relative to the `:imagesdir:` variable.
  This `:imagesdir:` has to keep its default value (current doc dir) so that translated images for this document are found.
- This causes two regressions:
     1. Github preview of documents won't find the images
     2. The legacy Makefile won't either
- Update icon paths in GSIK.adoc.

Caerbannog:
This failure is because Jenkins runs the legacy Makefile instead of the CMake scripts. What should I do?

nickoe:
Wait untill I get time to review it properly or ciampix or someone else have some input.

Caerbannog:
To be clear, @ciampix answered in #144 :
> so if you commit this you break the makefile build. If cmake is complete we can just get rid of it altogether but I think it is better to preserve it for now...

Is issue #63 the only thing to resolve before we can call the cmake build system complete?

PR #135, opened 2015-07-28
Update GSIK while reproducing the steps in a recent build

Caerbannog:
These changes could use a second look: some are for clarity, some are arbitrary, and there are things I could not reproduce and assumed were outdated.


nickoe:
Looks ok to me, I gues you just want me to merge it as is.

PR #138, opened 2015-07-29
adjust image size on pdf for Pl_Editor

kinichiro:
adjusted image displaying size on pdf.
take a look at Pl_Editor pdf, please.

- I used inline image macro `image:images/abc.png[width="NN%"]` in a table,
  since block image macro `image::images/abc.png[scaledwidth="NN%"]` doesn't work in a table.
- I changed some table settings (width, valign), too.
- I left all the icons now.

### DIFF
--- a/src/Pl_Editor/Pl_Editor.adoc
+++ b/src/Pl_Editor/Pl_Editor.adoc
@@ -131,7 +131,7 @@ dependent on the paper size**.
 [[reference-corners-and-coordinates]]
 === Reference corners and coordinates:
 
-image:images/en/page_property_1.png[]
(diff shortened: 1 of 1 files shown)


kinichiro:
We were talking about this PR at here https://github.com/ciampix/kicad-doc/issues/132#issuecomment-125995850
And, once, I will abandon this PR, because I couldn't find the alternatives for using table.

I was trying to use "Horizontal labeled list" with the images instead of table.
[17.3.1. Horizontal labeled list style](http://www.methods.co.nz/asciidoc/userguide.html#_labeled_lists)

But labeled list was difficult to use with the images as same as table.
I could not get good PDF output, after all.

PR #142, opened 2015-08-06
Update pcbnew intro and getting started in kicad intro

reportingsjr:
Updated both intros so they have accurate information.

I changed the installation part in getting started in kicad intro so it has a better structure (includes info about stable and unstable).

I think there is a LOT of fluff and unnecessary info that could be removed before the stable release to make the docs clearer. Lots of the pcbnew docs need to be rewritten to include info about GAL tools, change "module" to footprint, etc.


nickoe:
This looks good to me, but I will wait until tomorrow with merging, in case others have any comments.

nickoe:
Awaiting feedback on comment that I made on one of the commits.

kinichiro:
IMHO,
It is good to me, both docs updates.
I think roadmap should be in release notes rather than in reference manual intro.

ciampix:
@kinichiro is right. In general a roadmap is not material for a reference manual.

ciampix:
@reportingsjr as soon as you do as @nickoe suggested we can pull. Thanks!

PR #144, opened 2015-08-25
Remove warning about matching directory and project name

cpavlina:
It is not necessary for a project to be created in a directory with the same name. I'm not sure if it ever was. This commit rewrites the warning to urge the user to keep each project in a unique directory, without the warning that the names much match.

### DIFF
--- a/src/KiCad/KiCad.adoc
+++ b/src/KiCad/KiCad.adoc
@@ -303,8 +303,8 @@ recommended to create a project as follows:
   or “Create a new project from template” icon.
 
 [WARNING]
-It is _strongly_ recommended to use the same name for both project file
-and its directory.
+It is recommended to use a unique directory for each KiCad project.
+Do not combine multiple projects into a single directory.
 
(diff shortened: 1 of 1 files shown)


ciampix:
The makefile use the dir name as an indicator for doc name, see:
DOCNAME:= $(shell basename `pwd`)
so if you commit this you break the makefile build. If cmake is complete we can just get rid of it altogether but I think it is better to preserve it for now...

cpavlina:
I'm confused - did you look at the changes? I didn't change the build system, this is in the actual documentation talking about naming the directory of a KICad project.

nickoe:
I think @ciampix meant to comment on https://github.com/KiCad/kicad-doc/pull/134

I think the change is more accurate and will merge it.

ciampix:
Yes, sorry, I was reading and thinking to something else at the same time and those two curiously matched. I just have to stop replying email when too tired. I beg your pardon.

PR #146, opened 2015-08-27
Use git info for package version. Fixes #63

### LINKED ISSUE #63: CMake package version is always 1.0.0

BrianSidebotham:
Use Git information to generate the documentation package version.

Because of how Git generates tags we need to be careful when using them for the package version information. See the logic in the code to see some protection against using the wrong tags.

We need to use a schema. The schema used in this commit is that if the branch is master, the package version will always be master. We could add a commit hash to that if necessary, but I think that just complicates things.

When creating a branch we should use the Major.Minor format which aligns to the KiCad release the documentation is for such as 4.0, 5.0, etc.

The Patch field of a tag on a branch should be a triplet where the patch field of Major.Minor.Patch is the documentation version field. e.g 5.0.1, 5.0.2, etc.

A tag on a branch has priority for the package version information. If no tag is present the branch name is used. If no branch name can be detected (i.e. git was not found) unknown is used as the package version.


nickoe:
Nice, I will have a look at this ASAP.

ciampix:
Very good! Thanks!

PR #149, opened 2015-09-01
Clean images in GSIK and share them between docs

Caerbannog:
This is the same as PR #134, reopened after I closed it by mistake.


ciampix:
Hi @Caerbannog do you know if and how is possible to re-enable Jenkins test build for docs as this PR testify it used to be?

Caerbannog:
Your are right it was quite useful.
I think the right person to ask is @mangelajo 
It seems that @nickoe tried to fix it yesterday:
http://ci.kicad-pcb.org/view/KiCad%20docs/job/kicad-doc-testing-pull-request/204/

nickoe:
Hi, yes, I want to reenable it, but I would like try to enable it properly, it might prove difficult, so maybe I will just reenable it as it was before in the meantime.

ciampix:
ok thanks!

PR #151, opened 2015-09-02
modify Pcbnew_python_scripting.adoc

kinichiro:
- add `pydoc pcbnew`
- add section reference for "Footprint Wizards"
- fix sample scripts (indent, missing quote, etc.)
- remove section "Building KiCad with Scripting Support" since this is a reference manual.
- remove section "Scripting Support" since this is a reference manual.


ciampix:
Thanks!

PR #168, opened 2015-09-28
Fix parallel build issue ( Fixes #167 )

### LINKED ISSUE #167: Support parallel builds
The dependencies for each target seem to be incomplete, causing parallel builds to fail.

It would be great to have this fixed before KiCad 4.0 is released, as I'd like to integrate documentation compilation into the Debian package building process. Without parallel make support, build time is increased dramatically on a multi-core system.

BrianSidebotham:
I got interested in the parallel build issue and fixed it. The epub and pdf build targets have a race condition probably with some intermediary file that causes the build to fail in paralllel builds. I tested with -j8 to test the failure and also to test the solution which looks good.


BrianSidebotham:
Sorry, I also removed a useless file which was just hanging about

nickoe:
Ok, I see. thank you, I will merge it! Weeee.

PR #169, opened 2015-09-28
Make KICAD_DOC_PATH relative so that the packager can work okay

BrianSidebotham:
At the moment `make package` fails because `KICAD_DOC_PATH` is not a relative path. With this fix you can set `CMAKE_INSTALL_PREFIX` and run `make install` successfully as well as build the package target


nickoe:
Hmmm, ok. I will merge this, but I simply don't understand why cpack will not package what is placed in the install prefix.

PR #171, opened 2015-09-30
Proofreading of Eeschema docs

JamesTaji:
I did some proofreading of the Eeschema docs while I am learning to use the software.


nickoe:
Changes looks sane. Thank you very much for fixing these details. Lets hope you like kicad and want to try it all such that we can get some more proofreading done :P

PR #175, opened 2015-10-01
Sure up multiple ax2 runs for parallel jobs in the same folder (Fixes #167)

### LINKED ISSUE #167: Support parallel builds

BrianSidebotham:
I'm marking this as possibly fixing #167 as it tightens the dependencies for not running a2x in parallel in the same working folder.


nickoe:
Ok, it built fine with -j4 this time, I will merge this. It took ten minutes to build this time. Definitely an improvement.

Issue #166, opened 2015-09-27
eliminate multi byte char from adoc


kinichiro:
IMHO, adoc should not use multi byte char
if it is able to express by single byte char.

I think these could be eliminated from adoc.
- multi byte dash (– -> -)
- multi byte single quotation (’ and ‘ -> ')
- multi byte double quotation (“ and ” -> ")

And I also think these could be permitted.
- →
- …
- °

I listed up the adoc lines containing multi byte char here.
https://gist.github.com/kinichiro/9f3919491644e6f0f76d

How do you think about this ?

Thanks

nickoe:
What is the reason for this, even though it is called asciidoc, it supports UTF-8. A2x just needs the lang attribute set.

kinichiro:
Because some multi byte char seems give a bad influence on some editor.

For example, editing eeschema_general_top_toolbar.adoc with vi.
You can see multi byte '-' is displayed '?'.
![multi_byte_1](https://cloud.githubusercontent.com/assets/4927823/10122927/18a9dd1e-6564-11e5-862a-6d19d279fa54.png)

Editing pcbnew_routing.adoc with vi.
You can see some '?', but they should be single white space.
![multi_byte_2](https://cloud.githubusercontent.com/assets/4927823/10122930/4f160d8c-6564-11e5-8e70-a36fdcb9efd5.png)

I don't mean all multi byte char should be eliminated.
But, I think we should not use it as far as possible for avoiding troubles.

ciampix:
I agree 100%.

LordBlick:
Tool should be reported as buggy, not using std utf8…

nickoe:
@kinichiro That is simply because your terminal or font does not support it properly, please use a proper environment. Something that supports UTF-8. But that part is indeed a bit funny, but the problem is that someone who wrote that has inserted non breaking space. If you inspect the positions with a hex editor you will see it as C2 A0 in the pcbnew doc. And an en dash in the eeschema doc.

I suspect theese codes actually appeared from the odt to adoc conversion.

I think it does not hurt keeping them, but I guess it wont really hurt ot fix them either. if you feel like fixing these, I am ok with accepting a pull request on this.

I guess using the non multibyte chars for theese special chars will make it easier to debug if we find funky build errors too. But the last part is only a suspicion.

kinichiro:
@nickoe 
>I suspect theese codes actually appeared from the odt to adoc conversion.
I think so, too.
[...]

PR #178, opened 2015-10-04
eliminate multi byte double quotation from adoc

kinichiro:
eliminate multi byte double quotation for issue #166

I also would like to eliminate from 
eeschema_schematic_creation_and_editing.adoc,
but this causes build error of eeschema_pdf_pl,
so I didn't fix this file only.

I left °(degree) in these 2 files.
`\&deg;` seems work as °, but I couldn't tell it was right.
- eeschema_component_library_editor.adoc
- eeschema_hierarchical_schematics.adoc


kinichiro:
If you are OK, 
Could you merge this PR, please ?

ciampix:
Thanks @kinichiro !

PR #182, opened 2015-10-10
fix pcbnew, footprint editor section and description about keep out area

kinichiro:
this is a small fix for pcbnew.


ciampix:
Many thanks for your work, it is very appreciated!

kinichiro:
Thanks you too.

PR #195, opened 2015-10-28
French translation of GSIK by Pierre BENETEAU

Caerbannog:


nickoe:
Thanks.

kinichiro:
You should update po file by latest .adoc before starting translation work.
Artifact fr doc seems mixed French and English.

ciampix:
Done

kinichiro:
@ciampix thanks !

but, I thought each translators should update target po file by themselves
according to
https://github.com/KiCad/kicad-doc/blob/master/translation_instructions.adoc

message above was sent to translator.

ciampix:
Sorry, probably that doc is not clear. I described there the procedure to start a translation from scratch. But anyway I though that you would be glad if I help updating the po file for the translators. Have I done something wrong? Would you update your file yourself? Please let me know what you prefer and why and I will try to follow your preferences.

In any case that workflow use the old build mechanism that should vanish substituted by the new cmake based when it will be ready of course.

PR #197, opened 2015-10-29
modify eeschema png to indexed color

kinichiro:
exchange png files under eeschema/images/ and eeschema/images/en/ to indexed color.
size of eeschema pdf and epub will be reduced.
for example, in my environment, eeschema.pdf 3.5MB -> 1.9MB


ciampix:
Ok probably I should create a little script to do this automatically from time to time...
Anyway thanks!

PR #201, opened 2015-10-31
fix pcbnew description

kinichiro:
- update descriptions of pcbnew according to the latest software


ciampix:
Very good piece of work! Thanks.

PR #206, opened 2015-11-05
Fix gsik img

kinichiro:
This PR organizes GSiK images.

- delete unused png files under the images/ directory
- move all icons to under the images/icons/ directory
- replace icon files with formal png files
- fix paths of icon files in po files according to the path modification above (->images/icons/)


ciampix:
Great! Thanks!

PR #208, opened 2015-11-07
update ja.po

kinichiro:
- update ja.po of eeschema and pcbnew


ciampix:
Thanks!

kinichiro:
Thanks you, too !

PR #216, opened 2015-12-04
End of french translations of GSIK and huge fixes of fr.po

MarcusFecit:
Two typos in english version fixed and updated in other languages.

E ringrazio il tizio che ha traduto il it.po. Mi ha bene aiutato..


ciampix:
... that was me. Thanks to you too!

PR #219, opened 2015-12-05
Avoid using the network during documentation build

GyrosGeier:
This should stop the documentation build from trying to pull in external files during the build.


ciampix:
Thanks!

Issue #244, opened 2015-12-26
Confusion between hierachical pins and labels in Eeschema


MarcusFecit:
Section 6.6, "Connections - hierarchical pins", Manual placement.
This section is supposed to show the placement of a hierarchical pin, but the image and the attributes description are for the properties of a hierarchical label.

evanshultz:
Agreed. I submitted a pull request for eeschema_hierarchical_schematics.adoc, with the screenshot below replacing images/eeschema_hierarchical_label.png:
<img width="202" alt="eeschema_hierarchical_pin" src="https://cloud.githubusercontent.com/assets/1936989/16013565/c6a21788-3142-11e6-9ddf-bcddcc083afa.png">

evanshultz:
I believe this issue can now be closed.

ciampix:
Yes, surely.

PR #245, opened 2015-12-26
Eeschema fr translations: beginning of chapter 6

MarcusFecit:
Stopped by issue #244


ciampix:
I think that you should go on and translate. When will be corrections in the orginal source file you have to just correct/translate  the new sentences...

PR #248, opened 2015-12-28
update GSiK flowchart for japanese

kinichiro:
- update .dia file
- export .png file and shrink by index color


ciampix:
Thanks!

PR #251, opened 2015-12-29
fix png filename for GerbView

kinichiro:
- rename .png filename (ex. 3360000029F09D033B0.png -> gerbview_main_screen.png)


ciampix:
Your help is _very_ appreciated, thanks!

kinichiro:
Thanks for merging.
eeschema and GSiK have same issue, but now some peoples are translating them.
I would like to fix this same issue after they finish translating.

PR #252, opened 2015-12-29
Add German Translation for Eeschema

transistorgrab:
German translation for the current Eeschema documentation.
Only text was translated, not images for the German localization were included (the English ones are used).
On a side note:
The English documentation needs to be reworked though… there is much double information throughout it and the terminology is not consistent sometimes. At least at one point the software does not work as described in the documentation (hierarchy navigation).

Best regards.


ciampix:
You are right. Please be free to propose even small patches to improve docs. I will be eager to merge those improvements!

PR #264, opened 2016-01-05
Translate doc_writing_style_policy in Japanese

yoneken:


ciampix:
Sorry I did this without thinking well on it. This guide is for doc authors, not translators, so I do not see any reason to translate it... If you do not mind I am reverting it. If I got it wrong I will let you merge it later...

yoneken:
Ah.. OK.
No problem!

PR #259, opened 2016-01-04
German translation of GSIK

tijuca:
Please consider pulling my work for a translation of GSIK into German.
I added some additional screen shots with German locales to get relationship between the translated part and the respective screen shot.
Currently I have some trouble on chapter 4.1, point 16. I don't know why the translated string and the related screen shot isn't used.

This pull request would also close #239 if accepted.

Regards
Carsten

PR #268, opened 2016-01-05
Fixing not working German translation string in GSIK

tijuca:
I pick up the suggestions from the pull request #259 and fixed the de.po file for GSIK. I would like to see the helping commands for updating a *.po file somewhere documented, so anybody will find them later again.

Regards
Carsten


ciampix:
Thanks!

PR #270, opened 2016-01-06
fix png filename for GSiK

kinichiro:
- rename .png filename (ex. 10000000000000DD000000946E66C399.png -> images/gsik_myconn3_l.png)


ciampix:
Thanks, I already said so but ... your work is much appreciated!

PR #272, opened 2016-01-06
Take some comments on GSIK into account

Caerbannog:
https://forum.kicad.info/t/will-getting-started-tutorial-be-updated/1463/39?u=caer

TODO: keeping the numbering consistent is a PITA when adding detailed steps


ciampix:
Thanks!

PR #275, opened 2016-01-07
Fixing regressions in getting_started_in_kicad.adoc

tijuca:
With b89162c2f9a378e08130dbce1cdab9d7d9fefbff there are small regressions introduced, the numbering was interrupted  after that commit. Fixed.

Re formating the file getting_started_in_kicad.adoc to break up long lines after 78 characters. And depending on previous commits update of all po files in GSIK with finally updating the German translation.


ciampix:
Thanks!

PR #278, opened 2016-01-09
German translation of kicad

tijuca:
Hi,
I was working further and have translated 'kicad' into German with additional adding screenshots with German locales.
While doing translation I have found some small issues within kicad.adoc which I fixed up before working further. But there are just small nitpicking things. Let me know if there are problems.

Regards
Carsten


ciampix:
Thanks! Next time create two separate pull request... it simplify searching for problems and revert problematic commits...

PR #280, opened 2016-01-10
Beginning of french tanslations for CvPcb

MarcusFecit:


ciampix:
Thanks!

MarcusFecit:
Le 10/01/2016 20:46, Marco Ciampa a écrit :
>
> Merged #280 <https://github.com/KiCad/kicad-doc/pull/280>.
>

Grazie Marco,

Some english strings are missing in the cvpcb/fr.po like :

msgid "About KiCad document: https://github.com/KiCad/kicad-doc/issues"
...
...

If you have a simple way to check and correct cvpcb/fr.po according to
the english version it would be very helpful

e Buona notte..

-- 
Marc BERLIOUX

ciampix:
I can do it for you but I prefer not to avoid conflicts that may eventually arise. See this:

https://github.com/KiCad/kicad-doc/blob/master/translation_instructions.adoc

in particular what you have to do is this:

cd src/YOUR-CHOSEN-MANUAL
make -f ../../utils/old-build-scripts/Makefile po/NN.po

that in your case comes down to:

cd src/kicad
make -f ../../utils/old-build-scripts/Makefile po/fr.po

bye

PR #282, opened 2016-01-11
fix png filename for eeschema

kinichiro:
- rename .png filename (ex. 1000000000000134000001411D7B987C.png -> eeschema_popup_without_element.png)

I could build successfully after these changes on my local PC.
I also checked all languages (de, en, es, fr, it, ja and pl).


ciampix:
Many many thanks!

kinichiro:
Thanks merging :-)

PR #287, opened 2016-01-14
Added bit about footprints being spread out on import in GAL

reportingsjr:
I added this feature to KiCad ~4 days ago.


ciampix:
Thanks!

PR #289, opened 2016-01-15
French CvPcb, chapter 6

MarcusFecit:


ciampix:
Thanks!

kinichiro:
Hi, @MarcusFecit 
It seems that some numbers of '\n' between 'msgid' and 'msgstr' are different in cvpcb/po/fr.po.
Then, if you try to update po now, you will get errors like below.
<pre>
$ cd kicad-doc/src/cvpcb/
$ touch *.adoc
$ make -f ../../utils/old-build-scripts/Makefile po/fr.po
...
Updating po/fr.po:
........................... done.
po/fr.po:206: 'msgid' and 'msgstr' entries do not both end with '\n'
msgfmt: found 1 fatal error
po/fr.po: 118 translated messages, 142 untranslated messages.
</pre>

I would like to suggest to fix this issue.
Best Regards,

PR #295, opened 2016-01-16
Adding idf_exporter translated into german

tijuca:
I worked on translating the manual for the IDF Exporter into German. It's complete translated with additional screenshots from the GUI in German.

Regards
Carsten


ciampix:
Thanks!

PR #298, opened 2016-01-17
add spanish gerbview manual

AntonioMR:
I'm sorry for three commits for this but other files were modified when I built the html documents and I don't know another way to revert them than re-push


ciampix:
No problem, thanks...

PR #309, opened 2016-01-24
Pl_Editor, translate manual to German

tijuca:
Following my work for the Pl_Editor manual translated into German with also respective screenshots from the German UI. A local build of the various documentation media is working so far. There are some small grammar things to fix later.

Let me know if something is miss behaving.

Regards
Carsten


ciampix:
Thanks!

PR #313, opened 2016-02-04
Reorganize layer description, add missing layers

Caerbannog:
Here are improvements to the Layer section of the documention for Pcbnew.
They were prompted by the following question:
https://forum.kicad.info/t/looking-for-info-layer-names/2218

More information should be given for "F.Fab" and "Margin" layers if someone knows about them.


ciampix:
Thanks!

Issue #342, opened 2016-04-18
KiCad


LarryJoy:
[KiCad.pdf](https://github.com/KiCad/kicad-doc/files/222969/KiCad.pdf)

KiCAD April 9, 2016
Document comments and feedback
- [x] 2.2, 1 st paragraph, 2 nd line, “If an other default...” Should be “If another default...”

ok fixed by @Caerbannog 

- [x] Page 13, bottom of page, “Project from Template”, 2 nd line, near the end, “...will load that templates info.html...” Should be “...will load that template's info.html...”

ok

- [x] Page 14, last figure at bottom of page, “Templates path”. Should be “Template's path” unless you're talking about multiple templates, in which case I think it would be “templates' path”.

ok fixed by @Caerbannog

Caerbannog:
It will be easier next time if you just paste your text here instead of including a pdf :smiley:

PR #345, opened 2016-04-18
Typos that partially fix issue #342

Caerbannog:
- [x] 2.2, 1st paragraph, 2nd line, “If an other default...” Should be “If another default...”

- [x] Page 13, bottom of page, “Project from Template”, 2nd line, near the end, “...will load that templates
info.html...” Should be “...will load that template's info.html...”

- [ ] Page 14, last figure at bottom of page, “Templates path”. Should be “Template's path” unless you're
talking about multiple templates, in which case I think it would be “templates' path”.

This last item is about a screenshot so it will be fixed in the KiCad sources first.


ciampix:
Thanks!

PR #347, opened 2016-04-20
Used "ё" character instead of simplified version of it ("е"). Some gr…

baranovskiykonstantin:
…amatical corrections.


ciampix:
Thanks!

PR #351, opened 2016-05-05
Revert "fix cmake for update po files"

ciampix:
Reverts KiCad/kicad-doc#350

from the build dir

 cmake .. 

gives:
CMake Error at CMakeModules/KiCadDocumentation.cmake:103 (add_dependencies):
  Cannot add target-level dependencies to non-existent target "updatepo_it".

  The add_dependencies works for top-level logical targets created by the
  add_executable, add_library, or add_custom_target commands.  If you want to
  add file-level dependencies see the DEPENDS option of the add_custom_target
  and add_custom_command commands.
Call Stack (most recent call first):
  src/idf_exporter/CMakeLists.txt:8 (KiCadDocumentation)

### DIFF
--- a/CMakeModules/KiCadDocumentation.cmake
+++ b/CMakeModules/KiCadDocumentation.cmake
@@ -29,11 +29,6 @@ macro( KiCadDocumentation DOCNAME )
         list( APPEND DOCCHAPTERS "${CNAME}" )
     endforeach()
 
-    foreach( DOC ${DOCCHAPTERFILES} )
-        set( DOCCHAPTERMASTERS "${DOCCHAPTERMASTERS} -m ${DOC}" )
-    endforeach()
-    separate_arguments( DOCCHAPTERMASTERS )
-
     # If we're not building a specific language, glob all languages
     if( "${SINGLE_LANGUAGE}" STREQUAL "" )
         # Get a list of all po translation files so we know what languages can be built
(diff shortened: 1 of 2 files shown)


kinichiro:
@ciampix sorry for troubled you.
I tested my PR with cmake 3.5.1 on ubuntu16.04(x86_64) and it succeeded.
I checked with cmake 2.8.12, I've got the same error,
<pre>
...
CMake Error at src/CMakeLists.txt:22 (add_custom_target):
  add_custom_target called with invalid target name
  "updatepo_cvpcb//po//de.po".  Target names may not contain a slash.  Use
  ADD_CUSTOM_COMMAND to generate files.  Set CMAKE_BACKWARDS_COMPATIBILITY to
  2.2 or lower to skip this check.
...
</pre>

It seems that result of `file(GLOB)` command is different between the versions,
with 3.5.1 "updatepo_cvpcb/po/de.po",
but with 2.8.12 "updatepo_cvpcb//po//de.po" (slash supplied twice).

I found the way to fix this. In src/CMakeLists.txt, add '+' after '/' like this.
<pre>
...
     14     string( REGEX REPLACE ".*/+po/+([a-z][a-z]).po" "\\1" LANG "${PO}" )
...
</pre>

This change worked fine with both 2.8.12 and 3.5.1 on my environment.

ciampix:
No problem. "if you don't commit yourself, you'll never fail" it is an old adage. Prepare a PR and I'll be happy to merge it.

kinichiro:
Thanks. I submitted #353 .

PR #353, opened 2016-05-05
fix cmake for update po files

kinichiro:
- support multiple .adoc files
- generate po with using .adoc relative path
- add custom target "updatepo_\<LANGUAGE\>" to update LANGUAGE specific po files
- add custom target "updatepo_all" to update ALL po files


ciampix:
Thanks!

PR #360, opened 2016-05-09
[Russian] Getting started in KiCad: added an refinement to KiCad desc…

baranovskiykonstantin:
…ription, was corrected the pronoun 'вы'.


ciampix:
Thanks!

PR #374, opened 2016-06-13
A number of small changes for English readability

evanshultz:
There were some instances of incorrect English, which I corrected. In other cases, I made changes so that the text read better, cleaned up duplicate information, or re-worded the text to sound more natural and straightforward. Lastly, I took a look at the document at a high level to try and align formatting and usage so the document fits together. There is a bit more that could be done; the unnecessary use of commas, in particular, is something that I'd like to change. For me, this is a test to see if my changes are approved and accepted. If so, I'd be happy to continue reviewing the remaining KiCad documentation in a similar manner.


ciampix:
Well I agree with most of these suggestions but ... why not suggesting smaller patches? Now it is a "take all or nothing" and there are some small passages that I need to clarify (my English is scarce but I cannot give up some sort of checking...)

For example, you changed:

 > Schematic files, which do not contain the components themselves.

into

> Schematic files, which does not contain the components themselves.

but Schematic files is plural so I do not understand this because, as I see it, the Schematic files are the subject, or not?

Forgive me if I am asserting something that embarrassingly show my ignorance...

evanshultz:
You're correct. Smaller patches would be better, perhaps in a section-by-section manner? I apologize as I'm new to editing other projects.

For the specific example you mentioned, I should be embarrassed. It is wrong. There can be multiple schematics in a project when using hierarchy and I confused myself. I reviewed and submitted a second patch.

ciampix:
Anyway ... thanks a lot!

PR #377, opened 2016-06-15
Finish fixing issue #244

evanshultz:
Updated the image, but it looks overly large. This is a 1:1 screenshot from Eeschema, but perhaps it should be scaled down? Also, the eeschema_hierarchical_label.png image can, if not used elsewhere, be deleted.


ciampix:
Thanks!

Issue #304, opened 2016-01-20
Template Section - min requirements to make a template work


scottcher:
I recently tried creating a new template and found the instructions for doing so in the documentation site to have gaps that left me wondering why my template wasn't working as expected.

I suggest adding to the required files section a project file.

When I tried creating a template, I included the info.html file and icon.png file in the meta directory but only include the file I was intending to template from in the main template folder - in my case, the PCB file.  When I used the template, what I got was a new project file with a new schema and new PCB file and a copy of my intended PCB template file.

the solution was to include a project file in the template folder but that wasn't clear from the template instructions.

ciampix:
Thanks for the heads-up. Would you point me to the document/section you are referring to?
May you suggest a text or (best) a patch?
TIA

scottcher:
Sure.  The document section I'm referring to is http://docs.kicad-pcb.org/en/kicad.html#_using_templates but not the entire section.  This section is more about how to use templates and doesn't provide much details around creating them.

I can try to write something up that might help others avoid the problem I encountered - I'd be happy to contribute.

evanshultz:
This may have been fixed. I'm able to create a new template with only the required file (info.html), with required and optional files, and with all of the previous plus project files (*.pro, *.sch, etc.). In all cases my new project has a project file, schematic file, and PCB file. My version info:
Application: kicad
Version: 4.0.2-stable release build
wxWidgets: Version 3.0.2 (debug,wchar_t,compiler with C++ ABI 1009,GCC 5.2.0,wx containers,compatible with 2.8)
Platform: Windows 8 (build 9200), 64-bit edition, 64 bit, Little endian, wxMSW
Boost version: 1.57.0
         USE_WX_GRAPHICS_CONTEXT=OFF
         USE_WX_OVERLAY=OFF
         KICAD_SCRIPTING=ON
         KICAD_SCRIPTING_MODULES=ON
         KICAD_SCRIPTING_WXPYTHON=ON
         USE_FP_LIB_TABLE=HARD_CODED_ON
         BUILD_GITHUB_PLUGIN=ON

Also on my Win10 PC, I see the templates at C:\Program Files\KiCad\share\kicad\template. Can anyone else confirm that's correct for Windows?

evanshultz:
@scottcher can you reproduce this issue using the 4.0.2 stable release? Please paste the info from Help > Copy Version Information if you can.

[...]

PR #379, opened 2016-06-17
Fix after reading issue #304.

evanshultz:
While I can't reproduce that issue, it did point out how unclear this section was. A user will typically want to know how to use something before going deeper and learning how to create it themselves. So I re-arranged. And then cleaned up some text so that I believe this section is far more approachable. A sentence was added explicitly pointing out the user must create template files, which should hopefully help with the issue mentioned above.

### DIFF
--- a/src/kicad/kicad.adoc
+++ b/src/kicad/kicad.adoc
@@ -390,65 +390,92 @@ files, libraries, PCB, etc.
 |Rebuild and redraw the tree view, sometimes needed after a tree change.
 |=======================================================================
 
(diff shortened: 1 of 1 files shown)


ciampix:
Well I agree and it seems to me very good, thanks!

kinichiro:
@ciampix you just closed and not merge this PR. Is it OK ?

ciampix:
Oh my, you are right! Sorry!

ciampix:
Done, thanks!

PR #382, opened 2016-06-18
Update getting_started_in_kicad.adoc

evanshultz:
Removed unnecessary tables, corrected language a bit, mention the footprint editor with Pcbnew.

### DIFF
--- a/src/getting_started_in_kicad/getting_started_in_kicad.adoc
+++ b/src/getting_started_in_kicad/getting_started_in_kicad.adoc
@@ -63,7 +63,7 @@ tools:
 |Pl Editor |Page layout editor|+*.kicad_wks+
 |===================================
  
-*Note:* The file extension list is not complete and only contains a
+NOTE: The file extension list is not complete and only contains a
 subset of the files that KiCad supports. It is useful for the basic
 understanding of which files are used for each KiCad application.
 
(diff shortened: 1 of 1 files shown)


ciampix:
Please note that NOTE: , WARNINGS: etc. are special admonition paragraphs and keywords are all uppercase...

evanshultz:
OK. The style guide is... nonexistant and I felt that admonitions clashed too much with tables (about which users are explicitly warned in the AsciiDoc guide). Admonitions also aren't used in the KiCad manager docs, in favor of **Note:**, etc..

So they should go back to admonitions in order for this patch to be merged? Anything else that needs to be changed in this patch?

ciampix:
Well I reported this because I already fixed a patch of yours with such an admonition "corrected" by you that was correct before, just for make you aware, not for blocking an otherwise good patch. Hope you understand my intention, your work is very appreciated. There was an ongoing debate on a style guide that went nowhere because there are already some good ones (see here for an example: http://asciidoctor.org/docs/asciidoc-recommended-practices/ ) and because asciidoc limit authors on an already good-enough style on itself. If there are, here and there, other non-asciidoc-ized admonitions, I consider those errors. If you correct this patch (since, I cannot edit your patch) I will be more that happy to merge it. Thanks again.

evanshultz:
Patch updated. Apologies for that and I'll remember. Will also read the linked doc before doing any more work.

evanshultz:
With the admonitions added back to the latest patch above, is there anything else stopping it from being merged?

ciampix:
You deleted any reference to the unstable nightly builds. While I have not any particular preference on this regard, may I know what was your reasoning behind such a choice? Do you followed something similar done on other KiCad docs or was it a choice of yours?

evanshultz:
There's a single section above each OS section which mentions stable versus unstable and the differences. Rather than have the same text repeated three times, once for each OS, the existing text one time, that applied to all 3 OSes seemed quite sufficient.

ciampix:
I see, ok, thanks!

PR #383, opened 2016-06-18
Edited Overview section

evanshultz:
Changed section names to present participle form, moved forward annotation above backward since it always must come first, and removed the "-" where used in "workflow" to align the spelling


ciampix:
Thanks!

Issue #226, opened 2015-12-15
GSIK: shortcut keys could benefit from graphic


bobc:
Some users have found the specification of shortcuts confusing e.g. "pressing the Place component shortcut key (a). "

This could be improved by replacing the text with a graphic icon, e.g something like ![key top](http://attosoft.info/images/keyboard-keys/alt/wpclipart-a_s.png)
Some small and open source icons would need to be generated or found.

It might also be useful to have a summary of the shortcut keys used in the GSIK in an Appendix.

ciampix:
Please mind that shortcut keys are not the same for every language. Adding icons could be nice but we should supply all keys icons for every language that could possibly use those ...

bobc:
All the shortcut keys I looked at are the same for each language, so I don't understand your comment. However, there are a limited number of keys, I don't see a problem providing graphics for each language.

Can you tell me where the translation tables for shortcut keys are, I will try to generate icons programatically.

ciampix:
You are right of course. I was erroneously thinking about menu hotkeys...

kinichiro:
Hi,
Here are just some my thoughts.

- shortcut key graphic icon
  GSiK already has many program icons in it,
  and I feel we don't want to increase them any more.
  I would like to read the text, not viewing graphical icons.
  Instead of icons, how about using '[' and ']' with bold.
  e.g. "pressing the Place component shortcut key **[a]**. "

- summary of shortcut keys list
  shortcut keys can be redefined by users, so they are changed and customized.
  How about mentioning about '?' key more strong ?
  In the note of seciton 3.1 item 5 tells about '?' key, but it is weak.
  Adding new section, say, "2.3 Hotkeys" or something with screenshot
  will help user to notice how to get the list.
  IMHO, GSiK is not a reference manual but a guidance text, so it should describe "how to".

- shortcut key -> hotkey
  BTW, I think that "shortcut key" should be replaced with "hotkey".

Regards.

bobc:
Icons are bad? Really? Instead of making a box with ASCII-art, it is surely better to make a real graphic. I am happy to provide the graphics, and make a pull request.

If you put "[a]", you then need to add more text to explain that means type the "a" key, and type '[', 'a', ']' keys. That might be obvious to us, but anything that simplifies things for people who are newbies is surely a useful improvement.

I would suggest just summarising the hotkeys used in the tutorial, with a note that ? shows the current list. Obviously if the user has redefined keys, he will know that...

[...]

PR #389, opened 2016-06-20
Finish changes for issue #226

evanshultz:
Add new hotkey syntax, remove a now-unneccesary note about viewing the hotkey list, and change "shortcut keys" to "hotkeys". I would appreciate a review by @ciampix and @kinchiro and @nicknoe, if you're interested. I also fixed one admonition. Note that I haven't reviewed for clarity and do see opportunities to improve this document; right now all I'm doing to is addressing the above issue.

### DIFF
--- a/src/getting_started_in_kicad/getting_started_in_kicad.adoc
+++ b/src/getting_started_in_kicad/getting_started_in_kicad.adoc
@@ -378,7 +378,7 @@ image::images/component_history.png[Component history]
 14. In case you make a mistake and want to delete a component, right
     click on the component and click 'Delete Component'. This will remove
     the component from the schematic. Alternatively, you can hover over the
-    component you want to delete and press the Delete key.
+    component you want to delete and press [Delete].
 
 15. You can also duplicate a component already on your schematic sheet
     by hovering over it and pressing [c]. Click where you want to
@@ -692,11 +692,11 @@ connection. Let's see how to do it.
 1.  Let us suppose that you have three 4-pin connectors that you want
     to connect together pin to pin. Use the label option (press [l]
     ) to label pin 4 of the P4 part. Name this label 'a1'. Now
-    press the Insert key to have the same item automatically
+    press [Insert] to have the same item automatically
     added on the pin below pin 4 (PIN 3). Notice how the label is
     automatically renamed 'a2'.
 
-2.  Press the Insert key two more times. This key corresponds to the
+2.  Press [Insert] two more times. This key corresponds to the
     action 'Repeat last item' and it is an infinitely useful command
     that can make your life a lot easier.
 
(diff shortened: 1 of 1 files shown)


kinichiro:
Hi, @evanshultz 
Thanks for your great works and I think this fix is good.

This is my personal feelings, though,
For Insert and Delete hotkey, I would like to add brackets too,
like [Insert] and [Delete] since the convention note describes below;
`In this document, hotkeys are expressed with brackets like this: [a]`
How do you think about this ?
If just Insert and Delete is better than with brackets, I'm OK.

Best Regards,

evanshultz:
Let me first understand. I believe you feel that adding the brackets makes it more clear that a particular key needs to be pressed, so the user has the think less about the content and can more easily follow the Getting Started guide. I like that idea.

It's a good question. I searched Google for a bit with queries like ""getting started" "delete key"", ""getting started" "press delete"" and ""user guide" "delete key"", etc. I found the following nomenclatures:
-DELETE key
-Delete key
-"Delete" key
-'Delete' key
-<Delete> key

It seems special characters are stripped out, so I think my searching of this type was bogus as the brackets were being excluded and gave me false negatives that brackets are used.

I then tried to check out style guides and keyboard shortcuts, to see what the computer/OS vendors and document creation folks say. Other than all caps being out of style, I didn't find anything else helpful.

All told, I see little consensus and thus haven't developed a sense of "right" and "wrong". It seems any of the numerous ways can be equally as instructive to the reader. So with that in mind, I vote to wait a couple days for anyone else to chime in and, if not, add the brackets and move on from this issue. OK?

nickoe:
To me it is more important that it is easily readable, it it gets easier to read it with brackets around keys, then ok.

But I have to say that; "Edit by hovering over the pad and press the e key.", is fairly readable by itself. This wa just a made up example.

ciampix:
Well, it seems that this way to present key combination nicely fits an "experimental" feature of asciidoctor:

http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#ui-macros

Given that I see a transition from asciidoc (actual python executable asciidoc compiler) to asciidoctor (ruby modern asciidoc compiler - used by github for example) for our docs compilation scripts in the future, for a number of reasons, I agree to this (key combinations surrounded by brackets).

ciampix:
There is even a macro example that could be just right for us... 

https://gist.github.com/mojavelinux/4736691

let me do some experiments ... then I may have the need of a cmake expert :-)

kinichiro:
Yes, as you say in your comment, I just wanted to indicate the key is particular one.
And, Yes, I would like to wait some experiments of @ciampix .

evanshultz:
In the interim, @ciampix would you be willing to merge my patch? I'd like to continue working on this doc but am afraid of breaking it if I continue on to later sections. It seems to me that merging this patch is safe, and a new patch in the future with the cool button style will work nicely.

ciampix:
Yes of course, as soon as Travis ends the check. Thanks!

ciampix:
Thanks!

PR #390, opened 2016-06-21
Oops. Didn't correctly align the sections headers.

evanshultz:


ciampix:
Thanks for your precision!

PR #391, opened 2016-06-21
Change to implicit numbering in a few sections

evanshultz:
This seems easier than explicit numbering, and asciidoc is correctly restarting the numbers within each section. Why is explicit being used? Was this done before and it broke? If implicit is OK, let me know and I'll update the patch to do this for all of the document.

### DIFF
--- a/src/getting_started_in_kicad/getting_started_in_kicad.adoc
+++ b/src/getting_started_in_kicad/getting_started_in_kicad.adoc
@@ -702,18 +702,18 @@ A with some other sequential pins of component B. In this case you have
 two options: the labelling method we already saw or the use of a bus
 connection. Let's see how to do it.
 
-1.  Let us suppose that you have three 4-pin connectors that you want
+.  Let us suppose that you have three 4-pin connectors that you want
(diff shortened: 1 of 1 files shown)


kinichiro:
Hi,
Before moving to ADOC, KiCad documents were made by LibreOffice Writer.
That old GSiK had written in numbered sentences.
The text of current version of GSiK .adoc was created by copying from that old one.
That is why explicit numbering is used in .adoc files.

IMHO, explicit numbering has good readability before processing.
I sometimes modify the adoc file by editor, 
and explicit numbering helps me since it tells me where I am :-)

In practice, no one reads plane adoc file before processing,
except the editors and the translators.

I would like to remain the explicit numbering.

Best regards,

ciampix:
I agree. It is such a little burden to maintain and it is useful for doc writers. Furthermore the asciidoc compiler warns the user when the numbering do not correspond, such a test with implicit numbering does not happen so it is possible that a spurious carriage return can silently restart numbering. Thank you (indeed) but no thanks.

evanshultz:
Got it.

PR #392, opened 2016-06-21
Update menu item syntax

evanshultz:
Change to the more clear syntax seen in other parts of the doc (like GSIK). It's also much easier to discriminate from file paths.


ciampix:
Thanks!

Issue #210, opened 2015-11-18
bom2csv Plugin Error


boris81:
This is regarding *Getting Started in KiCad* document, page 19, step 56.  If done as specified, the *bom2csv* plugin will not work.  

The following code should be added in the note section:
>*sudo apt-get install xsltproc*

Once installed, then the plugin should be downloaded:
>wget https://raw.githubusercontent.com/KiCad/kicad-source-mirror/master/eeschema/plugins/bom2csv.xsl

ciampix:
Good catch! That could be certainly added for Debian/Linux or similar and I am already preparing a patch but I wonder how users can do with Mac OS X or Windows...

nickoe:
The xsl files are includes in the windows installer.
Den 18/11/2015 09.00 skrev "Marco Ciampa" <notifications@github.com>:

kinichiro:
On Windows, xsltproc.exe is placed under the directory C:\Program Files\KiCad\bin .
I checked this with kicad-product-4.0.0_rc2-x86_64.exe.
OS X, I have no idea since I don't have it.

ciampix:
So, in Linux, it could not be mentioned since it should be just a matter of package dependencies that should already manage to have xsltproc installed with KiCad, right?
If the Mac OS X already contains the xlstproc executable, this note could be avoided at all...

Caerbannog:
@ciampix Linux package dependencies should indeed install it. But someone will probably have the problem one day: bad package, install from source, ...

kinichiro:
IMHO, how about mentioning that xsltproc is the OSS and users might install by themselves.
Like description about Freerouter in GSiK.
And add the installation example for Ubuntu only.

nickoe:
Does this need some changes to the docs? I will soon tag the docs for the release, so any changed that needs to be included shall be made very very soon.

boris81:
@nickoe Please add the following information on the document:

>*sudo apt-get install xsltproc*

It should go on page 19, step 56, note section, right before *wget https://raw.githubusercontent.com/KiCad/kicad-source-mirror/master/eeschema/plugins/bom2csv.*

evanshultz:
I'd like to update the doc and close this issue, but I'm not sure what to do. Can someone help?

It seems to me this is broken down by OS.

*Linux*
Since this will (apparently) be handled with a package dependency, do we or do we not need to include the code above? Maybe it's best to note this dependency and include the code, so that we've provided all the important info for users if they have a problem?

[...]

PR #395, opened 2016-06-22
Issue #210

evanshultz:
Note explicit method to get BOM dependencies under Linux and add section for Windows.

### DIFF
--- a/src/getting_started_in_kicad/getting_started_in_kicad.adoc
+++ b/src/getting_started_in_kicad/getting_started_in_kicad.adoc
@@ -636,12 +636,24 @@ easily editable or scriptable.
 +
 [NOTE]
 =====================================================================
-The *.xsl file is located in __plugins__ directory of the KiCad
-installation, which is located at: /usr/lib/kicad/plugins/.
+*Windows:*
 
-Or get the file via:
+xsltproc.exe is located at _<KiCad install directory>\bin_, while the included
+*.xsl files can be found at _<KiCad install directory>\bin\scripting\plugins_.
+
+*Linux:*
+
+The *.xsl files are located in the _plugins_ directory of the KiCad installation,
+which is located at: _/usr/lib/kicad/plugins/_.
+
+You can get bom2csv.xsl via:
 ---------
 wget https://raw.githubusercontent.com/KiCad/kicad-source-mirror/master/eeschema/plugins/bom2csv.xsl
+---------
+
+If xsltproc is missing, you can downloaded it with:
+---------
+sudo apt-get install xsltproc
 ---------
 =====================================================================
 +
--- a/src/getting_started_in_kicad/getting_started_in_kicad.adoc
+++ b/src/getting_started_in_kicad/getting_started_in_kicad.adoc
@@ -647,17 +647,18 @@ xsl files are located at: _/usr/lib/kicad/plugins/_.
 
 *Apple OS X:*
 
-If you need xsltproc, use the following to download and install it:
+If xsltproc is missing, you can download and install it with:
 ---------
 brew install libxslt
 ---------
 
-You will find xsl files at: _/Library/Application Support/kicad/plugins/_.
+xsl files are located at: _/Library/Application Support/kicad/plugins/_.
 
 *Windows:*
 
-xsltproc.exe and the included xsl files will be installed at
-_\bin_ and _\bin\scripting\plugins_, respectively.
+xsltproc.exe and the included xsl files will be located at
+_＜KiCad install directory＞\bin_ and
+_＜KiCad install directory＞\bin\scripting\plugins_, respectively.
 
 *All platforms:*
 
--- a/src/getting_started_in_kicad/getting_started_in_kicad.adoc
+++ b/src/getting_started_in_kicad/getting_started_in_kicad.adoc
@@ -636,25 +636,27 @@ easily editable or scriptable.
 +
 [NOTE]
 =====================================================================
-*Windows:*
-
-xsltproc.exe is located at _<KiCad install directory>\bin_, while the included
-*.xsl files can be found at _<KiCad install directory>\bin\scripting\plugins_.
-
 *Linux:*
 
-The *.xsl files are located in the _plugins_ directory of the KiCad installation,
-which is located at: _/usr/lib/kicad/plugins/_.
+*.xsl files are located at: _/usr/lib/kicad/plugins/_.
 
-You can get bom2csv.xsl via:
+If these files are missing, they are available via the following
+commands (change the *.xsl file name as desired):
 ---------
+sudo apt-get install xsltproc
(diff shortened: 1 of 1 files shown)


kinichiro:
Hi,

This sentense `If xsltproc is missing, you can downloaded it with:`
`..., you can download and install it with:` is better (at least 'downloaded' is not good).

Can we add OSX here, can't we ?
I don't have Mac and I don't know if xsltproc is installed on OSX or not.
(@adamwolf I appreciate if you could tell about this)
*.xsl files might be located at `/Library/Application Support/kicad/plugins/`
according to this investigation.
https://forum.kicad.info/t/rc2-package-directory-structures-across-the-platforms/1664/5

And I would like to suggest about description orders,
first xsltproc, next plugins directory,
Linux, OSX and Windows (according to "Downloading and installing KiCad" section)

Best regards,

ciampix:
@kinichiro you took the words right out of my keyboard ... :-)

adamwolf:
I spent a day or two (!!!) trying to get the doc build chain working on OS
X about a year ago, and it was not easy.  Currently, I use a Linux VM just
to build the docs.  It might be easier now--if someone on OS X could try,
that would be great.

(Is that what you were asking?)

nickoe:
@adamwolf Why don't just use the prebuilt packages as found on http://docs.kicad-pcb.org/ for the latest and http://downloads.kicad-pcb.org/docs/ for the stable?

adamwolf:
Because I had built the linux VM before the prebuilt were up and working,
honestly.  No real reason.

ciampix:
To @adamwolf, reference: (Is that what you were asking?) I am afraid, no. Please see the @kinichiro post about xsltproc ... this is about BOM creation stylesheets ... Thanks.

adamwolf:
Sorry.  xsltproc is included in OS X dev tools, I think, but you can install a newer version if you use brew with just

brew install libxslt

This probably needs someone without OS X dev tools installed to confirm if it comes with the OS or with the dev tools.

kinichiro:
Thanks for your clarification.

evanshultz:
Patch updated with feedback from the first two comments and to avoid so many "located at" so close together. Linux and OS X are unfinished as I don't know the correct xsltproc paths (assume installed by default) and I don't have a way to check. Please review, and thank you.

kinichiro:
Hi,

I had read through the part you added.
I don't think we need to describe where xsltproc is installed on Linux and OSX.
We just need to how to install it.
On OSX, "KiCad install directory" is not needed.
On Windows section, 'and' should be needed.
We don't want put '*' or '.' character in front of line since these are the special chars in asciidoc.

How do you like this description, below ?
<pre>
*Linux:*

If xsltproc is missing, you can downloaded and install it with:
---------
sudo apt-get install xsltproc
---------

xsl files are located at: _/usr/lib/kicad/plugins/_.

*Apple OS X:*

If xsltproc is missing, you can downloaded and install it with:
---------
brew install libxslt
---------

xsl files are located at: _/Library/Application Support/kicad/plugins/_.

*Windows:*

xsltproc.exe can be found at: _<KiCad install directory>\bin_.

xsl files are located at: _<KiCad install directory>\bin\scripting\plugins_.

*All platforms:*

You can get the latest bom2csv.xsl via:

https://raw.githubusercontent.com/KiCad/kicad-source-mirror/master/eeschema/plugins/bom2csv.xsl
</pre>

This results PDF like this.
gsik_xsltproc

Best regards,

evanshultz:
I prefer not to repeat the same sentences for each OS and would thus like the formatting I used in 6fc36be better, but breaking it up this way is OK with me. However, @boris81 is saying it's good to go as-is. I'd like @ciampix and @nicknoe to take a look; I can modify the patch to whatever the consensus turns out to be.

kinichiro:
OK, I will withdraw my idea.

But could you check this in Windows section, please ?
`xsltproc.exe the included *.xsl files can be found at`
Does this need `and`, doesn't it ?

And Apple OS X section,
- Could you add `brew install libxslt` to indicate how to install xsltproc.
- I think ＜KiCad install directory＞ is not needed.

Thanks.

ciampix:
mmm I still prefer @kinichiro version... and under MacOSX I remember that someone stated that xsltproc is included int the Apple XCode dev code tools... could some Apple user confirm?

kinichiro:
>someone stated that xsltproc is included int the Apple XCode ...

Adam said about this in comment above, here https://github.com/KiCad/kicad-doc/pull/395#issuecomment-227794125

evanshultz:
OK. Great. I just want to serve the users, and close open issues, so if further changes are needed I'm happy to do it. My interest was in getting more eyes on the ideas to build up a consensus. I'll update the patch later and then let's see if everybody is happy.

And thanks for the sharp eyes to point out my mistakes, @kinichiro!

evanshultz:
Updated the patch. Is this fine? I still think Windows needs ＜KiCad install directory＞ to be clear about where to find the files. Maybe we could prepend the paths with Program Files\KiCad? (I only use Windows so I don't know about other OSes.)

kinichiro:
> I still think Windows needs ＜KiCad install directory＞ to be clear about where to find the files.

I agree. Windows needs ＜KiCad install directory＞.
Windows has 2 versions (64bit and 32bit) of KiCad,
and ＜KiCad install directory＞ depends on user's choice of installation.

evanshultz:
OK, great. Patch updated. Anything else?

Note that when I started this patch some other patches to this doc hadn't yet been merged. @ciampix please be careful to check that only Eeschema item #58 is affected. Thanks!

kinichiro:
Thanks for your contribution. :-)

ciampix:
Sorry for being late, busy these days... just a couple of things:

First:

> apt-get install xsltproc 

it's a bit not-distribution-agnostic. If you agree I would add "... for Debian derivated distros like Ubuntu, yumm install xsltproc for RedHat and derivated or whatever it needs to install xsltproc package on your specific distro.

Second:

> "Circuit board board and footprint editor"

is this repetition correct? To me it seems a bit weird...

kinichiro:
@ciampix I agree, and 'yumm' -> 'yum' is right.

ciampix:
Please have a look on how it goes now...

kinichiro:
I think it is ok, and especially good for linux newbie users.
Thanks.

ciampix:
And for "Circuit board board and footprint editor" @evanshultz, please have a say!

evanshultz:
A couple thoughts on this patch:
1. To me, it's more clear to put the distro above the code used. That way, the user knows if the code applies to them before reaching it, rather than having to go back once they realize they just passed the code for them.
2. How about a link to Xcode so it's a bit easier? Should be https://developer.apple.com/xcode/. That allows removing the "tool from the Apple site that should contain it," text since the user will see the blue underlined text.

@ciampix Yes, one "board" is sufficient. Probably my fault, and thanks for checking. PR submitted.

PR #397, opened 2016-06-23
Polish GerbView doc

evanshultz:
Since this is a small doc I did it in one go. Hope that's OK.

General polish to the text and simplification where possible. Used "RS-274X" since that's the exact nomenclature used in the Gerber standard. Removed explicit references to Pcbnew's output, since GerbView can view files from most any PCB design tool. Made all toolbar icon descriptions verb-noun. Fixed table so the last left toolbar icon isn't hidden. Removed subjective or unhelpful text like "Previously (a long time ago) it was also possible to load so called Dcodes, but those are now obsolete and is therefore not possible anymore.". Pushed the printing section to the end so that moving items is already described.

NOTE: As I moved the display modes into new sections ("==="), the images disappeared; reverting this change didn't fix it. I'll take a look at this more later, but I'm submitting this patch for review since I think everything might be OK and the issue is GitHub trying to render the differences.

### DIFF
--- a/src/gerbview/gerbview.adoc
+++ b/src/gerbview/gerbview.adoc
@@ -51,9 +51,13 @@ viewer. Up to 32 files can be displayed at once.
 For more information about the Gerber file format please read
 http://www.ucamco.com/files/downloads/file/81/the_gerber_file_format_specification.pdf[The Gerber File Format Specification].
 
-== Main window
+== Interface
+
+=== Main window
 
 image::images/gerbview_main_screen.png[scaledwidth="95%",alt="gerbview_main_screen_png"]
+
+<<<<<
 
 === Top toolbar
 
@@ -94,6 +98,8 @@ image::images/gerbview_top_toolbar.png[scaledwidth="95%",alt="gerbview_top_toolb
 |image:images/gerbview_top_info.png[width="70%",alt="gerbview_top_info_png"]
 |Display info about active layer
 |=======================================================================
+
+<<<<<
 
 === Left toolbar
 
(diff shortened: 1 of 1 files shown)


kinichiro:
(1) contents structure

I took a look at your new structure of GerbView.
Is this the structure what you want to re-organaize, right ?

```
1 Introduction to GerbView
2 Main window
  2.1 Top toolbar
  2.2 Left toolbar
3 Commands in menu bar
  3.1 File menu
  3.2 Preferences menu
  3.3 Miscellaneous menu
4 Layers Manager
5 Display modes
  5.1 Raw mode
  5.2 Stacked mode
  5.3 Transparency mode
  5.4 Layer occlusion
6 Moving items
7 Printing
```

Original structure is below.

```
1  Introduction to GerbView
2  Main Screen
3  Top toolbar
4  Left toolbar
5  Commands in menu bar
  5.1 File menu
    5.1.1  Export to Pcbnew
  5.2 Preferences menu
  5.3 Miscellaneous menu
6  Layer Manager
  6.1 Modes to show Gerber layers
  6.2 Effect of layer selection for drawings
7  Print layers
  7.1 Print dialog access
  7.2 Move block command
```

(2) png filename appears in Contents

This expression (section header with image tag)

```
=== Raw mode image:images/icons/gbr_select_mode0.png[gbr_select_mode0_png]
```

causes displaying png filename in Contents, like this.
gerbview_pdf2

It is not so big problem, but I would like to avoid this.
Now, I have no idea to avoid this issue.
Do you know the resolution for this ?

(3) page break

If you build your changes into pdf, asciidoc layouts "Left toolbar" section like this.
gerbview_pdf

At this time, an asciidoc can not layout the table like above well ...
Don't you like add page break (`<<<<<`) before "Left toolbar", like this.

```
     96 |=======================================================================
     97
     98 <<<<<
     99
    100 === Left toolbar
    101
    102 [width="100%",cols="10%,5%,85%",]
```

To tell the truth, I don't like using page breaks for formatting the pages, I don't feel it is a good way.
Though this is a workaround for this problem of asciidoc, now.

Best regards,

evanshultz:
Thanks for the review! I've been hoping for feedback so this can get merged.

1. Yes, it was, except I realized that the Layers Manager is part of the interface. So I've moved that section with the others and retitled the section.

2. I moved the icons within text under each section header.

3. Great idea. I didn't create PDFs and so didn't see. Page breaks are now added before the two big tables. Are there other places where this should happen?

Please see the latest patch and let me know if you find other changes needed. Note that I used "some" computers in the Display modes section rather than "older" computers as drivers or another issue the documentation may not want to specify could be at fault. Is this OK?

kinichiro:
Thanks !

PR #401, opened 2016-06-27
A few small updates after seeing other docs

evanshultz:
Removed the not-very-descriptive "and a lot more" from the list of files, removed bold from the file/folder table where it didn't add value, made all standalone file names bold to help them stand out, changed the wildcard in filenames to match the existing standard in other KiCad documentation.

### DIFF
--- a/src/kicad/kicad.adoc
+++ b/src/kicad/kicad.adoc
@@ -301,7 +301,7 @@ schematic editor containing a backup of the components used in the schematic.
 
 == Using KiCad manager
 
-The KiCad Manager (KiCad or kicad.exe) is a tool which can easily run the other tools
+The KiCad Manager (kicad or kicad.exe) is a tool which can easily run the other tools
 (schematic and PCB editors, Gerber viewer and utility tools) when creating a design.
 
 Running the other tools from KiCad manager has some advantages:
--- a/src/kicad/kicad.adoc
+++ b/src/kicad/kicad.adoc
@@ -70,12 +70,11 @@ the successful development and maintenance of complex electronic boards.
 KiCad does not present any board-size limitation and it can
 handle up to 32 copper layers, 14 technical layers and 4 auxiliary layers.
 
-KiCad can create all the files necessary for building printed boards:
+KiCad can create all the files necessary for building printed circuit boards, including:
 
 * Gerber files for photo-plotters
 * drilling files
 * component location files
-* and a lot more.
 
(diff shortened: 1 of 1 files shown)


ciampix:
== Using KiCad manager
 
-The KiCad Manager (kicad or kicad.exe file) is a tool which can easily run the other tools
+The KiCad Manager (KiCad or kicad.exe) is a tool which can easily run the other tools

this is wrong. Under Linux or Mac kicad (command) is all lowercase...

evanshultz:
Oh! My mistake. I though it was qualifying the shorthand name of "KiCad" without using the full "KiCad manager". I didn't realize it was the command names in various OSes. Updated patch. Please let me know if there's anything else to be changed.

evanshultz:
Any progress reviewing and/or merging this patch? I'm a bit worried that I need an escape character ("\") on line 297, so `example.\*` instead of `example.*`.

kinichiro:
>I'm a bit worried that I need an escape character ("\") on line 297,
 so example.\* instead of example.*.

It works fine without escape, like below.
kicad_example

To confirm if escape is needed or not, you need to build docs by yourself.
https://github.com/KiCad/kicad-doc/blob/master/README.adoc describes how to built.
ubuntu 16.04 on VMWare is enough if you are using Windows PC.

evanshultz:
@kinichiro 
Thanks for the review and verification. Are there more issues stopping this PR from being merged? The biggest change was making file names bold, which I saw in other documentation. But if that's not something we want, then we can include it in the style guide. :)

I am using Windows, and will read about building the docs. So far I've just viewed on GitHub. (Yes, I know...)

kinichiro:
I've read through all your changes and I don't think there are any other issues.
I just thought to keep this PR as open 2 or 3 days to see if someone gives comments to this PR.

The way of emphasizing text is different between asciidoc and github markdown.
See http://asciidoctor.org/docs/user-manual/#a-basic-comparison-by-example

IMHO, for emphasizing text, I prefer to use bold ('*') rather than italic ('_').
I feel bold is better to read, my personal feelings, though.

evanshultz:
That's great. Thanks! Will wait for additional feedback.

I prefer bold as well for readability.

ciampix:
I too prefer bold. Italic is more useful in non technical texts, to indicate something peculiar or ironic or so...

kinichiro:
Thanks for your great works.

PR #404, opened 2016-06-29
Update plugins.adoc

evanshultz:
Very minor editing to improve clarity (some may be subjective, like making file names bold, but it matches the other KiCad documentation I've seen). Gave the two tutorial sections simple but descriptive names instead of nondescript numbers. Fixed a few typos. Removed "shall" which was jarring and out-of-place with the plebian writing style.

Questions:
1. On line 539 I assume there's only one dialog for selection files, but maybe that's wrong?
2. There are references which don't work. Are these in another format and need to be changed to asciidoc? Does anyone know?
3. Different formatting for file names, classes, functions, etc. Maybe this is good, or maybe there's already a standard I haven't found?

### DIFF
--- a/src/plugins/plugins.adoc
+++ b/src/plugins/plugins.adoc
@@ -47,7 +47,7 @@ Published on January 29, 2016.
 The KiCad plugin system is a framework for extending the capabilities
 of KiCad using shared libraries. One of the main advantages of using
 a plugin is that it is not necessary to rebuild the KiCad suite while
-developing a plugin; in fact plugins can be built with the aid of a
+developing a plugin; in fact, plugins can be built with the aid of a
 very small set of headers from the KiCad source tree. Removing the
 requirement to build KiCad during plugin development greatly increases
 productivity by ensuring that the developer only compiles code directly
@@ -56,9 +56,9 @@ time required for each build and test cycle.
 
 Plugins were initially developed for the 3D model viewer to make it
 possible to support more types of 3D models without requiring major
-changes to the KiCad source for each new model supported. The plugin
-framework was later generalized so that in the future developers can
-create different classes of plugins.  Currently only 3D plugins are
+changes to the KiCad source for each new model type supported. The
+plugin framework was later generalized so that in the future developers
+can create different classes of plugins.  Currently only 3D plugins are
 implemented within KiCad but it is envisioned that a PCB plugin will
 eventually be developed to make it possible for users to implement
 data Importers and Exporters.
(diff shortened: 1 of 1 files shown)


kinichiro:
Hi,

>1 . On line 539 I assume there's only one dialog for selection files, but maybe that's wrong?

Do you mean here ?
@cbernardo Any thoughts about this question and this PR ?

```
...
    538 `wrl` in turn until a plugin returns visualization data. The
    539 file filters provided by each plugin are passed onto the 3D
    540 file selector dialogs to improve the browsing UI.
...
```

>2 . There are references which don't work.

I believe all the 2 references are working right.

```
...
described in <<REF:PLUGIN_CLASSES,Plugin Classes>>. The API is implemented
...
<<REF:CLASS_PLUGIN_3D, Plugin Class: PLUGIN_3D>>. The corresponding
...
```

These 2 REFs are translated to link in PDF, like below.
plugins_2_1
and
plugins_2_2
Why do you think references doesn't work ?
What is the problem you mentioned above ?

>3 . Different formatting for file names, classes, functions, etc.

Do you add '*' for those names like this ? I think it is not good.

```
...
*.s3d_plugin_demo1.cpp*
[source,c]
-----
#include <iostream>
...
```

This causes unwanted heading '.' like this.
plugins_3_1
In asciidoc, the heading '.' text following source code is a "Callouts".
See http://www.methods.co.nz/asciidoc/userguide.html "20. Callouts".

Best regards,

cbernardo:
In general I think it's OK except for (a) addition of unnecessary commas, (b) change of headings (why are they being changed?), and (c) breaking the formatting by placing '*' around strings.

Regarding L540, we can use 'dialog' rather than 'dialogs' since there is really only one file selector dialog but I was counting the instance in 2 different Footprint Properties dialogs.

evanshultz:
Thanks all. Please see below and provide additional feedback.

@kinichiro
1. OK. "dialog" is is.

2. The link style didn't match all the other KiCad docs I've seen. See the updated patch where I converted to the style I've seen elsewhere using lower case names separate by hypens and implicit links instead of "REF". (This also works in GitHub, where the original version doesn't.) Change back, or leave like this?

3. Reverted.

@cbernardo 
a. Which commas are unnecessary? I see commas I've removed but I must have missed any added ones.

b. As in the patch comment above, numbers have no meaning but to show order. Using "basic" and "advanced" is more descriptive, though hardly enlightening. I feel it's better for a tutorial when users won't have experience with KiCad plugins before. Also, most of the previous KiCad documentation uses section titles everywhere, even if they're not linked. (I'll happily create them if requested.) But, certainly, if everyone dislikes it I will revert.

c. Reverted above code blocks as mentioned above. Is this what you meant about broken formatting?

In normal text, I saw in some places that code blocks were used around file names and elsewhere there was no formatting. I was trying to standardize and the other KiCad docs use bold for file names. I've changed all to code blocks, as was used many places in the original document. But for consistency among all KiCad documentation, I prefer bold.

kinichiro:
@evanshultz

>2 . There are references which don't work.

OK. I think your fixes about references are good.

>3 . Different formatting for file names, classes, functions, etc.

OK.

evanshultz:
@kinichiro 
Great! Thanks for checking.

@cbernardo 
Can you please reply to my comment above? If there's anything still unsuitable, please let me know. If all is well, then this PR can be merged. Thank you.

evanshultz:
@kinichiro @ciampix With no objections in over a month, can this be merged?

cbernardo:
No objections here.

PR #406, opened 2016-07-07
Update gerbview.adoc

evanshultz:
Add link to Excellon format; removed text in Layers Manager section that's already covered better in the Display Modes section.


ciampix:
Thanks!

PR #418, opened 2016-09-10
small fixup for gui translation howto and adding German translation

tijuca:
Hi,

I have done some initial work on translating the gui_translation_howto into German. The translation is mostly complete but the screenshots from the original English source are a little bit mixed up from different languages and OS systems, also seems a little bit outdated. So I skipped to create new localized screenshots. Also the Bazaar parts in the English manual needs to be adopted to the actual Git status quo first.

While working on the German translation I realized some small issues in the English original text like not uniformed usage of the word PoEdit and usage of http URLs there https instead is possible. So I changed those issues first and afterwards going further with the German translation.

If something is not clear please let me know.


kinichiro:
Thanks

PR #420, opened 2016-09-18
De de

tijuca:
Hi,

I made some changes to the cvpcb documentation which I think are useful. There are no big modifications, it's more fitting into typical text width and adding some more explanatory HTML 'alt' tags.
I prepared on top of this changes also a update for the German translation, but first I'd like to see the changes to cvpcb*.adoc accepted (or not).

### DIFF
--- a/src/cvpcb/cvpcb.adoc
+++ b/src/cvpcb/cvpcb.adoc
@@ -12,9 +12,9 @@ _Reference manual_
 
 This document is Copyright (C) 2010-2015 by it's contributors as listed
 below. You may distribute it and/or modify it under the terms of either
-the GNU General Public License (http://www.gnu.org/licenses/gpl.html),
+the GNU General Public License (https://www.gnu.org/licenses/gpl.html),
 version 3 or later, or the Creative Commons Attribution License
-(http://creativecommons.org/licenses/by/3.0/), version 3.0 or later.
+(https://creativecommons.org/licenses/by/3.0/), version 3.0 or later.
 
 All trademarks within this guide belong to their legitimate owners.
 
--- a/src/cvpcb/cvpcb.adoc
+++ b/src/cvpcb/cvpcb.adoc
@@ -108,7 +108,7 @@ assignment via equivalence files.
 
 *CvPcb is only invoked from the schematic capture program Eeschema*, by the tool:
 
-image:images/icons/run-cvpcb.png[run cvpcb]
+image:images/icons/run-cvpcb.png[Calling CvPcb]
 
 Eeschema automatically passes the correct data (component list and footprints) to CvPcb.
 There is no update to do (unless some new components are not yet annotated), just run Cvpcb.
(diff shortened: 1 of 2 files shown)


nickoe:
I have no hard feelings for this, but is it really necessary do things like this? Don't this just mark other translations as outdated while they are not, or will poedit not call this a fuzzy translation?

ciampix:
Please do not apply this patch. IMHO if someone is willing to do some trivial or cosmetic changes in this way he/she should take care of also updating _all_ languages po strings and unfuzzy them (as I did for some typos many times...) and doing this operation be sure to warn about all translators to let them update their translations before doing this to avoid nasty conflicts that may arise from this operation.

tijuca:
@ciampix 
So how you want to update such files at all? Only by a few people? Most of the docu stuff are more or less outdated since development is going further and further? What are updates in the end? With **every** change there will surely be fuzzy strings in the end.
For classical source code development there are some common rules for contributions, like also in KiCad, why this isn't made for the documentation as well?

I've seen and read such arguments like written from @ciampix  in some other projects. I disagree here completely. No, it's not my job to update **all** other language files as well, but it's normally my interest in a free software project to going further and help the project itself in places there I can. And I see no problem at all to do such changes in a development workflow until kicad-doc is reaching a new RC state. The translations are done by the the persons which take care for their language. And it's typical to get some conflicts while updating the files. Where is the difference between kicad-doc and kicad-i18n stuff? I can't see it or in other words, there is no for me.

So for me would this be the end of contribution on kicad-*as I have obviously some other kind of view. I won't work on files that have not really up to date chapters as this work seems wrong and useless to me. I've spend many hours in researching the source files and writing hopefully good translations, But with answers like above I'm not really motivated to do more.
I miss some kind of plan to create up to date documentation for KiCad.

PS: GitHub is the wrong place fur such discussions, they should be done on the ML as there will be reached more people who have some opinion on such topics.

nickoe:
I would encourage people to stop being offend for no reason.

Did you try to test if this gives fuzzy translations? Ideally it should not, but I don't know. I suspect it could mark them as fuzzy.

If it does not, then this should not be a problem. If it does, thwn I thin it would be better to update the po-files also, such as to not waste translators time re-reviewing strings that does not really need to.

ciampix:
@tijuca I think (no offend here please, I really do _not_ mean to offend you) you are missing the point here.
What I meant is that if you look at your patch more closely even with the github web interface you will see that your patch changes many strings formatting in such a way that it really do no change the text but it will provoke the marking of fuzzy strings in many already translated messages creating _a lot_ of work for the translators. I understand that you are not working on the stable branch of docs but I would suggest to try to change only if there is really a reason for doing it and not for cosmetic reasons of the asciidoc source that will be lost in the pdf or html conversion anyway.

kinichiro:
Hi,

Unneeded changes of line break brings invalidating translations.
I downloaded branch de_DE and `make updatepo_ja` and confirmed.
PO4A appears to recognize these as fuzzy, changing of sentence.

This PR is sent to master (development) branch,
and I believe that the improvements of docs often might need to break of old translations,
though these changes of line breaks bring no improvements.
I would not like to accept these kind of changes.

About image alt captions, I think those are improvements.
Those are good. We could not do that jobs while maintain 4.0 docs.

I would like to recommend @tijuca to include only improvements of image alt captions to your PR.

Thanks.

tijuca:
I update my PR and skipped the text formatting as requested. I won't change anything in those files in the future as I have a different view on what is an improvement and want to spend my energy on things with more outcome for me.
If this PR would be accepted I can readjust a already prepared update for the German translation of CvPcb.

ciampix:
Very good and many thanks for your understanding. Yes, sometimes to contribute to a free software project you have to have a lot of patience!

PR #421, opened 2016-09-25
Add documentation for the new 3D features.

cbernardo:
This adds documentation for the new features provided by the 3D plugin system and the new 3DViewer. Some images may already be outdated since they were produced before the new 3DViewer was merged.


nickoe:
Thanks!

PR #423, opened 2016-10-01
Update and fixups for German translation of CvPvb for stable branch

tijuca:
As done before for current master this PR is containing a update for the German translation of CvPcb for the stable branch of kicad-doc.


ciampix:
Thanks!

PR #425, opened 2016-10-15
Update and fixups for German translation of Eeschema, small fixup CvPcb

tijuca:
Rework of the whole German Eeschema translation. Adding some localized screenshots for the documentation of Eeschema. Finally a really small fix for the addendum.de file. 

I plan to adopt the adjustments into the 4.0 branch asap, but that will be less work than the first preparations.


ciampix:
Thanks!

PR #437, opened 2016-10-23
Latest updates for Russian translation.

baranovskiykonstantin:
Translations of manuals was updated for idf_exporter and pl_editor from stable branch.


kinichiro:
Thanks for your contribution :-)

PR #439, opened 2016-10-27
Change "OS X" to "macOS"

evanshultz:
Apple changes their desktop OS name, so it seems to me we should follow suit. The main KiCad website (http://kicad-pcb.org/download/osx/) hasn't updated their terminology, though.


ciampix:
Thanks for your attention to details!

PR #442, opened 2016-11-10
De: Update of Gerbview manual

tijuca:
Update of the current German Gerbview manual against the master branch.


ciampix:
Thanks!

PR #458, opened 2016-12-10
[pcbnew] Fix 15.1 and 15.5 descriptions

kinichiro:
- SEGVIA was renamed VIA since a long time.
- board.GetSegZones() must be removed from doc:
  this is a very old method only used to read very old boards.


src/pcbnew/pcbnew_python_scripting.adoc
```diff
@@ -207,10 +206,10 @@ rendered, and you have some parameters you can edit.
 
 If the plugins are not properly distributed to your system package,
 you can find the latest versions in the KiCad source tree at
-link:http://bazaar.launchpad.net/~kicad-product-committers/kicad/product/files/head:/pcbnew/scripting/plugins/[launchpad].
+link:https://git.launchpad.net/kicad/tree/pcbnew/python/plugins[launchpad].
 
 They should be located in for example `C:\Program
-Files\KiCad\bin\scripting\plugins`.
+Files\KiCad\share\kicad\scripting\plugins`.
```
nickoe: I don't think the change in this line is correct. On windows the install target is indeed bin\scripting\plugins as far as I can see. This is where the xsl files are located.

nickoe: Ok, share/kicad/scripting/plugins is correct. At least that is where cmake installs them to.

PR #460, opened 2016-12-11
[pcbnew] Fix 15.7 python example code

kinichiro:
- just copy from /usr/share/kicad/scripting/plugins/FPC_(SMD_type)_footprintwizard.py


kinichiro:
This 'GetLineTickness' typo seems fixed in master branch of source code.
But, KiCad version 4.0.x still has this typo in it.
So, This PR should be applied to only 4.0 branch.
Then I will close this PR.

PR #470, opened 2016-12-14
Update German translation for upcomming version 4.0.5

tijuca:
I've updated and reviewed the existing *.po files on the current German translations for the KiCad manuals. Pcbnew could get some more love but currently there are no fuzzy strings any in all files more.


kinichiro:
Now I have confirmed below since this is a big PR  ;-)

- This PR is targeted for 4.0 branch
- all updated files are restricted only for German translation
- travis-ci build was OK

tijuca:
It looks big, yes. The biggest change is happen to Eeschema due the added screenshots. But most changes in this PR are reflected by adding translations to string that doesn't needed normal a translation like words e.g. 'TEXT or 'Via', used units that are the same in German, or just author names. I decide to do that because the output of 'make updatepo_de' was a little bit misleading. Other parts are long standing here ...

kinichiro:
Great works. Thanks !

PR #473, opened 2017-01-19
Add description and image of manual routing

johnbeard:


kinichiro:
Hi,
I don't think src/pcbnew/images/Pcbnew_routing_posture.svg file is needed.
Only src/pcbnew/images/Pcbnew_routing_posture.png is used to build document.

johnbeard:
The SVG is the original source of the image - the PNG is just the render. That way if someone wants to modify the image, they dont have to re-draw the SVG? The other vector graphics for the staggering diagrams are done in the same way.

nickoe:
This looks good to me.

ciampix:
Very good!

PR #474, opened 2017-01-23
Improve confusing heading "Hierarchy creation of headlines"

bobc:
The heading "Hierarchy creation of headlines" doesn't make sense, I suggest "Summary of hierarchy creation" is more understandable.


ciampix:
Absolutely, thanks!

PR #475, opened 2017-01-31
Action plugin doc

jsreynaud:
Requested documentation for new feature: Able to run a python script from a menu directly from pcbnew.


kinichiro:
Thanks for your KiCad doc contribution.
I think it is good and I will merge this.
For build breakage, I will small fix after merging this.

PR #478, opened 2017-02-16
Add lang-id.conf

kinichiro:


kinichiro:
Add lang-id.conf for temporary workaround aginst build break.
We can see the travis-ci test build log from https://travis-ci.org/KiCad/kicad-doc/builds/202271964 .

This lang-id.conf is now copy of lang-en.conf.
In the future lang-id.conf should be translated to Indonesian
and should be posted as PR to https://github.com/asciidoc/asciidoc .

nickoe:
I would rather we commit the indonesian in both master and 4.0.

kinichiro:
Both master and 4.0, agree.
There is no Indonesian lang-id.conf yet, see  https://github.com/asciidoc/asciidoc .

PR #482, opened 2017-02-21
editorial of src/kicad/po/de.po

Mechtilde:
Hello,
I did a editorial review for the German file src/kicad/po/de.po
Kind regards
Mechtilde


ciampix:
Thanks!

PR #483, opened 2017-02-22
Completed GSIK Indonesian translation

triyanwn:
Indonesian translation of Getting Started in KiCad is now complete.


ciampix:
Thanks!

PR #488, opened 2017-03-23
Update_ja_20170321

starfort-jp:
Po files catch up with the latest.
Update images.


ciampix:
Thanks!

starfort-jp:
Thank you for your quick response. :)

PR #492, opened 2017-04-04
Rewrite of Pcbnew footprint management

johnbeard:
Rewrite the Pcbnew footprint management documentation to (I hope) explain how the library system in Pcbnew (and CvPcb) works, with more details about the plugins and more use of paragraphs and admonition blocks to avoid large dense blocks of explanation.

Since CvPcb is becoming less important and this section was obviously copied between Pcbnew and CvPcb, I moved it all into Pcbnew and removed the CvPcb section. I don't know if it's possible to link between the two, however.

I'm not quite sure if I've done this all correctly, as I'm sure it will be a pain for translators. I copied the fp wizard images over to Pcbnew, but haven't updated the po files.


stambaughw:
On 4/4/2017 9:28 AM, John Beard wrote:
> Rewrite the Pcbnew footprint management documentation to (I hope)
> explain how the library system in Pcbnew (and CvPcb) works, with more
> details about the plugins and more use of paragraphs and admonition
> blocks to avoid large dense blocks of explanation.
> 
> Since CvPcb is becoming less important and this section was obviously
> copied between Pcbnew and CvPcb, I moved it all into Pcbnew and removed
> the CvPcb section. I don't know if it's possible to link between the
> two, however.

I wrote and copied the original docs because this was written back when
our docs were odt files.  There was no easy way to create a separate doc
for the footprint library manager section and link it to both the CvPcb
and Pcbnew docs.  I asked about doing something similar now that we are
using asciidoc but I never got an answer and I don't have time to look
into it myself.  Ideally, this would be the way to do it so we don't
have to update the same information in two place.  Having common doc
sections is definitely something that would be useful as we have other
features that span multiple applications.  We should continue to include
this information in the CvPcb doc until (if?) it goes away.

johnbeard:
OK, the commit to remove the CvPCb stuff is separate. This now just updates the Pcbnew documentation.

ciampix:
You added even the Italian screenshots! Great work, thanks!

ciampix:
Sorry, forgot to merge... :-)

ciampix:
To @stambaughw, the thing is: there is no way to create an "universal" external link. I mean that if you refer to an external piece of documentation in an asciidoc, you can only create a link, and that link is an URL that points to html or an on line PDF or an on line epub but AFAIK there is no way to point to the same format in which the doc is created, that is html if the doc is in html format, pdf if it is pdf and so on... perhaps using some obscure macro/scripting feature that I bet would be different and incompatible in each asciidoc compiler implementation we will use in the future...

ciampix:
I am going to post the issue in the asciidoc(tor) developers mailing list anyway...

johnbeard:
@ciampix - Perhaps if there was some sort of "common" directory holding .adoc files that all the targets (pcbnew, cvpcb, etc) could "include"? I don't know if asciidoc/tor can specify some addition "include directory" (crudely speaking, a bit like `-I` for GCC or `PYTHONPATH` for Python). 

Asciidoctor _can_ have the "root" dir set from the command line: "The base directory defaults to the directory of the main document and can be overridden from the CLI or API" [[1](https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/include-directive.adoc)].

If that all fails, you could always write CMake to copy .adoc files from the hypothetical `src/common` to each of `src/pcbnew` and `src/cvpcb`.

The question then is what to do when two targets have almost identical, but not quite identical, documentation. For example, the pcbnew footprint lib project is relative to the PCB file, but the cvpcb one is relative to the netlist.

ciampix:
@johnbeard: yes the include directive is already there in the asciidoc specs, and it can avoid writing things more that one time. But I was referencing to _links_ or _references_ to external doc without including it to avoid having the same things written in more than one place. If the contents differ slightly then this is not a real problem since we are not really writing the same things so there is not a waste of space... though it could be disorienting having to collect things disseminated in different manuals... sorry I am a bit puzzled now...

johnbeard:
@ciampix I was thinking of ways to "include the same thing twice", rather than "link to resources outside this target".

One problem with the external link is that it introduces a hard inter-target dependency on the layout of the documentation in the final location. The location of the Pcbnew docs must be fixed relative to the Cvpcb link to it to work. If that's not an issue, I guess Cmake could set asciidoc(tor) variables/attributes (if that's a thing), or even do replacements using sed or something to write in the correct link locations for the given target before compiling the document.

It would be nice to be able to link rather than include the same thing twice, then you can say "see the linked Pcbnew docs, but Cvpcb used the netlist location, not the PCB file location". That saves copy-pasting (in this case) about 3000 words of manual and then modifying a couple of words. That results in bit-rotting as the two diverge slowly, as well as adding a lot of translation overhead. The extra space taken by duplicating text and images is secondary, I think.

ciampix:
Well we can avoid all these problems grouping all docs in a big manual. We can keep the "Getting started..." and che cheat-sheet separated and keep some or all other docs in one source dir. In this way we obtain a better reuse of screen-shots and icons even in those OS that does not understand symlinks (Windows)... what do you think of this? I suggest to move this discussion on mailing list though...

stambaughw:
Marco,

Thanks for the information.  It certainly would be useful if you could
include or insert a separate page into a document.  I am surprised that
asciidoc doesn't support something like this.

Cheers,

Wayne

stambaughw:
I guess I should have read further along in this thread.  It seems to me
that include would do the correct thing as long as we are keeping the
documents separate.  This would eliminate the need to maintain the same
documentation in two different places.

The other option is to create one KiCad user's manual rather than
breaking it into separate manuals and make the footprint management
section which can be linked anywhere in the manual.  This would be a
significant deviation from how our documentation has been done so we may
want to get feedback before we head down that path.

ciampix:
Wayne,
sorry my English is very bad. I meant that you certainly can include another asciidoc document from an asciidoc document (we are actually doing exactly this right now) as you may see here:

http://www.methods.co.nz/asciidoc/userguide.html#_processing_document_sections_separately

What you can't (easily) do is cross referencing a document or part of it from another in a document agnostic way as we can do with html hyperlinks. I mean that you can do it (you can use URLS) but it will not work for each document type as there is no such a concept in PDF for example, nor in epub. So after having though on this, asciidoc can't resolve in itself what appears to be a limit of the output format and not something to be blamed of the tool in itself...

ciampix:
Grouping the manuals although can be very effective in this regards since cross-references always works as certainly works including sections more that one time...

ciampix:
Wayne, sorry I was writing during your posting...

stambaughw:
Marco,

No need to apologize for your English.  I understand.  In the short
term, it may be prudent to move the footprint management section into a
common folder and include it in the CvPcb and Pcbnew documents.  If we
ever move to a single document solution, we can rearrange the file
layout at that time.  I know from experience, if we copy and paste the
footprint management section, they will diverge at some point in the
future.  It already has happened in the existing documents.  Including
it should eliminate that issue.

Cheers,

Wayne

PR #505, opened 2017-09-24
Pcbnew: Fix for action plugin example.

baranovskiykonstantin:


nickoe:
I assume this is correct, and works with master. Thanks.

PR #509, opened 2017-10-03
Remove incorrect note about ratsnest tooltip

### LINKED ISSUE #508: "The tool-tip is backwards" note is incorrect
`getting_started_in_kicad.adoc` contains a note (line 814):

    NOTE: The tool-tip is backwards; pressing this button
    actually displays the ratsnest.

This is not true - instead, the tooltip is dynamic and reflects the action that will take place when you press the button. It changes from "show" to "hide" and _vice versa_. In both 4.0 and master branches the behavior of the tooltip is correct, if a bit clunky.

cpavlina:
Fixes #508


ciampix:
Thanks!

PR #521, opened 2017-12-03
[Eeschema] Added option description of BOM plugin

baranovskiykonstantin:
On MS Windows operating systems, BOM generator dialog has special option
that controls visibility of external plugin window.
This commit adds description for it in Eeschema manual.


ciampix:
Thanks!

PR #533, opened 2018-02-08
Eeschema update (chapters 1-3)

orsonmmz:


orsonmmz:
The build fails for languages other than English due to some icons that were obsolete and are removed now. Should I temporarily restore them to keep the documentation successfully build?

ciampix:
Not needed... I'll pull it anyway and fix it in a moment... thanks!

ciampix:
Ops I think I misunderstood you... if you delete an image you should take care of deleting any reference to it in the source also. If you do not do this, apart the compilation issue, docs are not coherent!
I can fix the compilation issue commenting these image references out but you definitely should check for consistency of the docs!

ciampix:
We should definitely create:
1) make check_docrootname (for checking missing links and such)
2) make updatepo_docrootname (for updating just the single document po files)
in cmake!

orsonmmz:
Sorry for the problem. What is the recommended way then? I modify only .adoc files, should I modify .po files to remove any text referencing to the image or just the image?

ciampix:
It's not your fault. We should have some sanity check in cmake before let someone ask for a pull.
For now just look for the FIXME in eeschema source tree with a simple:
 grep "FIXME" *
rebase and fix it, thanks!

orsonmmz:
Ok, now I understood the problem. I am suprised asciidoc has not reported any issues here, I will be more careful now.

ciampix:
When we will switch to asciidoctor there will be some warning during test compilation...

Anyway many thanks!

PR #534, opened 2018-02-09
Eeschema: fixed image references

orsonmmz:
- Modified a reference to library management dialog to refer to
corresponding chapter in the manual.
- Removed description of hierarchy navigation tool,
as it does not exist anymore.
- Removed description of options that most likely were accessible on
the top toolbar, but they are not there anymore.


ciampix:
Thanks!

PR #540, opened 2018-02-24
Update README.adoc

tijuca:
Just a few adjustments to the README.adoc file.
Basically I stumbled around a failed build of the Russian PDFs files before I found the reason due a missed package.


ciampix:
Thanks!

PR #542, opened 2018-02-25
README.adoc: sort also second Debian/Ubuntu package list

tijuca:
This is a supplement for d14a73f4. Sorting the second list also
alphabetical and adding the needed font package 'fonts-freefont-ttf' to
the list too.
Fix a small sorting issue an the list above to keep entries really
alphabetical.


ciampix:
Thanks!

PR #549, opened 2018-04-07
Update part of src/eeschema/po/ja.po from poedit id 657 to 839

kinichiro:
* Update part of src/eeschema/po/ja.po from poedit id 657 to 839

* Correct previous updating about src/eeschema/po/ja.po

by @asukiaaa


ciampix:
Thanks!

Issue #548, opened 2018-04-07
Icons exist in the document icon folder, but not found in source code


kinichiro:
These icon files are existing in src/(document)/images/icons/ folder.
But those are not found in source code bitmaps_png/sources/ as *.svg.
Now, I have no idea why those are missing.
Some of them might be just changed file name (or icon name).
Any of them might be removed since that functions were obsolete.

To solve this issue, these steps would be required
 - need to find these .png in .adoc and find the right icon name,
 - copy that icon in src/(document)/images/icons/ and remove old one,
 - update .adoc and *.po

**Cvpcb**
auto_association.png	-> auto_associe.png
coordination_system.png	-> polar_coord.png
copy_to_clipboard.png	-> copy.png
filter_component.png	-> module_filtered_list.png
filter_library.png	-> module_library_list.png
filter_pincount.png	-> module_pin_filtered_list.png
open_3d_model_view.png	-> shape_3d.png
pointer_style.png	-> cursor_shape.png
rotate_back_x.png	-> rotate_neg_x.png
rotate_back_y.png	-> rotate_neg_y.png
rotate_back_z.png	-> rotate_neg_z.png
rotate_forw_x.png	-> rotate_pos_x.png
rotate_forw_y.png	-> rotate_pos_y.png
rotate_forw_z.png	-> rotate_pos_z.png
run-cvpcb.png		-> cvpcb.png
toggle_edge_sketch.png	-> show_mod_edge.png
toggle_grid.png		-> grid.png
toggle_ortho.png	-> ortho.png
toggle_pads_sketch.png	-> pad_sketch.png
toggle_text_sketch.png	-> text_sketch.png

**Eeschema**
(none)

**GerbView**
(none)

**Getting Started in KiCad**
add_text.png  -> text.png

**KiCad**
(none)

**Pcbnew**
add_mires.png		-> add_pcb_target.png
add_text.png		-> text.png
array_line.png		-> array.png
array_module.png	-> array.png
array_pad.png		-> array.png
array_target.png	-> array.png
array_text.png		-> array.png
array_zone.png		-> array.png
duplicate_footprint.png	-> unused
duplicate_line.png	-> duplicate.png
duplicate_module.png	-> duplicate.png
duplicate_pad.png	-> duplicate.png
duplicate_target.png	-> duplicate.png
duplicate_text.png	-> duplicate.png
duplicate_zone.png	-> duplicate.png
new.png			-> new_board.png

**PL_Editor**
[...]

PR #550, opened 2018-04-07
Remove unused icons in GSiK and Pcbnew

kinichiro:
This PR resolves the part of #548 .
- removed images/icons/*.png those are not used in .adoc files
- left new.png, since it was not used in .adoc, but used in po/*.po, still


ciampix:
Great, thanks!

kinichiro:
Thanks.

PR #555, opened 2018-04-08
Fixed PL_Editor icons

kinichiro:
This fixes part of #548 for PL_Editor.
- remove old icons
- add new icons
- update adoc and po files to use new icocns


ciampix:
Thanks for this and for all others!

PR #556, opened 2018-04-09
 Update src/eeschema/po/ja.po from poedit id 877 to 1380 for 5.0.0

asukiaaa:
Update Japanese translation file for eeschema.


ciampix:
Thanks!

asukiaaa:
Thank you!

Issue #73, opened 2015-06-09
cmake build always rebuilds all


cpavlina:
When building using a cmake-generated Makefile, make rebuilds the whole kit and caboodle, even if only one file was changed since the last make. This is quite slow on some systems!

nickoe:
Mentioning @BrianSidebotham in case he is not notified and he wants to help fix this issue.

BrianSidebotham:
Thanks @nickoe I had missed this.

cpavlina, I agree - the dependency list is not complete. This is something to be completed, it's not that straight forward with CMake as far as I remember because of the custom commands, but it certainly should be doable.

The things I can recommend for you are:

1. make defaults to the `all` target and so will clearly rebuild all. If you're only editing one document, you can simply just make that target instead of all. e.g. `make CvPcb`, etc.
2. Configure with [SINGLE_LANGUAGE](https://github.com/ciampix/kicad-doc/blob/master/cmake-readme.adoc#single_language) flag

onitake:
There's another implication to this: If you use `make install` to copy the built documentation into a staging directory, it will also rebuild everything from scratch. That can be slightly annoying.

BrianSidebotham:
In order to fix this problem we're going to have to move from `add_custom_target()` to `add_custom_command()`

The change will be a bit intrusive and will probably require a bit of testing before it's right so I suggest we deal with this problem after a stable release. Things are working okay at the moment even if they are a little annoying.

This should be fixed though, it's not a quick build process to build all the docs!

nickoe:
Thatis where theSI

onitake:
Thanks @BrianSidebotham, these are exactly my thoughts on the topic.

PR #566, opened 2018-05-04
Cleanup cmake rules

StefanBruens:
Currently, the whole documentation gets build twice when doing "make && make install", as all files are created using commands specified in "add_custom_target".

As documented for [add_custom_target](https://cmake.org/cmake/help/v3.6/command/add_custom_target.html#command:add_custom_target), these commands are always executed. The correct way to only rebuild on changes is a combination of `add_custom_target(target DEPENDS file)` and `add_custom_command(OUTPUT file COMMANDS ...)`.

### DIFF
--- a/CMakeModules/AsciidocHelpers.cmake
+++ b/CMakeModules/AsciidocHelpers.cmake
@@ -46,12 +46,18 @@ macro( add_adoc_pdf_target TARGET INFILE OUTFILE LANGUAGE )
         #list( APPEND _A2X_OPTIONS -D "${_OUTPUT_DIR}" )
     endif()
 
-    add_custom_target( ${TARGET} ALL ${A2X_COMMAND} ${_A2X_OPTIONS} ${INFILE} )
+    add_custom_target( ${TARGET} DEPENDS ${OUTFILE} )
+    add_custom_command( OUTPUT ${OUTFILE}
+                        COMMAND ${A2X_COMMAND} ${_A2X_OPTIONS} ${INFILE}
+                        DEPENDS ${INFILE} )
 endmacro()
 
 # Add an asciidoc to HTML conversion target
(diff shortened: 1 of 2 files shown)


StefanBruens:
Fixes #73 
Don't know why the build fails here, a local build works fine

nickoe:
Did you try a clean build?

ciampix:
I just cloned your repo and confirmed that it works as expected... don't know what is going on here ... I'm for applying it ... @nickoe ?

ciampix:
to @nickoe: are we sure that Travis do correctly clean up before building?

nickoe:
Yes, I am pretty sure. that it the intention. Are you building the same way as travis?

`mkdir build && cd build && cmake .. && make -j4`

ciampix:
Yes definitely

ciampix:
May I go ahead and merge?

StefanBruens:
I also use the same command sequence to build

StefanBruens:
Just had an idea whats going wrong here - I only build html and pdf, but not epub.
Both epub and pdf use an intermediate adoc -> docbook step, and executing both in parallel likely corrupts the docbook xml file.

a) add a dependency to force serialization
b) add an additional rule to create the docbook file, and feed the docbook file into a2x (explicitly supported by a2x)

ciampix:
Ok try commit and lets see what it happens ... BTW the above command build all docs in parallel and I didn't see any problem at all...

StefanBruens:
Can you just wait a little bit - I have fixed it locally and just need to cleanup

ciampix:
Great! Thanks @StefanBruens!

nickoe:
Now the build time went from 2 hr 9 min to about 43 min. Thanks.

nickoe:
@StefanBruens  Will you be able to throw some tips on https://github.com/KiCad/kicad-i18n/issues/208 ?

Issue #155, opened 2015-09-14
CMake: SINGLE_LANGUAGE is not very flexible


BrianSidebotham:
The **SINGLE_LANGUAGE** option is not very flexible. It would be better to have an option that is a list of languages which defaults to all languages.

So, I propose we remove the **SINGLE_LANGUAGE** option and instead have a **LANGUAGES** option. To most people building a single language this will just mean reconfiguring with `-DLANGUAGES=en` instead of `-DSINGLE_LANGUAGE=en` but it does mean that any number of languages could be built, or excluded from a build. e.g `-DLANGUAGES=en;fr`

I see this as a pretty low-priority task at the moment, but it would be good to tidy the option up.

nickoe:
Yes, indeed, I think a LANGUAGES option is more flexible.

ciampix:
Agree

nickoe:
Fixed by pr #567

PR #567, opened 2018-05-05
Add support for selecting multiple languages

StefanBruens:
Previously, either all available languages could be built, or a single
language could be selected using e.g. -DSINGLE_LANGUAGE=de. Add LANGUAGES,
which accepts a list of languages, e.g. -DLANGUAGES=en;de;fr.
SINGLE_LANGUAGE is kept as an alias to LANGUAGES for backwards
compatibility.

See issue #155.

Signed-off-by: Stefan Brüns <stefan.bruens@rwth-aachen.de>

### DIFF
--- a/CMakeModules/KiCadDocumentation.cmake
+++ b/CMakeModules/KiCadDocumentation.cmake
@@ -5,6 +5,16 @@
 # (c)2015 Brian Sidebotham <brian.sidebotham@gmail.com>
 #
 
+macro( INTERSECTION out_list list1 list2)
+# Store the intersection between the two given lists in var_name.
+    foreach( L ${list1} )
+        if( "${list2}" MATCHES "(^|;)${L}(;|$)" )
+            list( APPEND intersect_tmp ${L} )
+        endif()
+    endforeach()
+    set( ${out_list} ${intersect_tmp})
+endmacro(INTERSECTION)
+
 macro( KiCadDocumentation DOCNAME )
 
     # Add the cvpcb documentation targets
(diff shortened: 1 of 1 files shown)


nickoe:
Looks good to me. Thank you for fixing these issues.

ciampix:
First of all thanks for this.

I was just wondering if it feasible, and if you think that it could make sense, without too much effort, to add the different languages as compilation targets and getting rid of all these variable options (or keeping then... it shouldn't hurt to have more ways to do the same thing...).

I mean enabling something like this:

```
make all (already make all formats for all documents)
make document-name_format-name_all (missing)
make document-name_all_all (missing)
```

nickoe:
What do you mean by "name"? Is that the language code? IMHO I think it is better to stick with the cmake variable to defined the languages compiled.

ciampix:
Examples:
```
eeschema_all_all
eeschema_html_all
```
and I forgot (sorry):
``make document-name_all_language``
example:
`` make eeschema_all_it``

nickoe:
Honestly, I think we should invest more time in trying to make it work with asciidoctor to see if it builds faster than asciidoc. Compilation at the moment is quite slow and I have  the impression that asciidoctor could be faster, although we probably need to drop the epub format.

ciampix:
I could agree. I wish I was able to fix cmake script by myself though... anyway it is not necessarily true that we would have to drop the epub format: ascidoctor-epub3 works fairly well once you have formatted text in a suitable way, that means: master index document without any text content but made only of include directives...

StefanBruens:
regarding `<docname>_all_all` -  I think this is covered by the `<docname>` target. See `make help` for a list of available targets.

ciampix:
Yes definitely. And creating html it is from 10 to 30 time faster (and results are IMHO prettier...)

ciampix:
there are 3 asciidoctor:
* [asciidoctor](https://asciidoctor.org/) (plain) that creates html 
* [asciidoctor-pdf](https://asciidoctor.org/docs/asciidoctor-pdf/) that creates (directly) pdf
* [asciidoctor-epub3](https://asciidoctor.org/docs/asciidoctor-epub3/) that creates epub3

PR #568, opened 2018-05-10
Update packages to install for Debian/Ubuntu

tijuca:
The old list of needed packages isn't reflecting the current state for
Debian Stable (Stretch) / Testing and Ubuntu 18.04 as well.
For Spanish documentation there is also texlive-lang-spanish needed, if
texlive-lang-all is used. texlive-lang-dutch isn't existing in current
Debian based distributions so remove this package from the list and use
the successor package texlive-lang-european.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -48,21 +48,21 @@ package.
 
 === Debian / Ubuntu
 
-To install the dependencies on Debian / Ubuntu  run the following (requires
+To install the dependencies on Debian / Ubuntu run the following (requires
 about 1.5GiB of space):
 
     sudo apt-get install asciidoc cmake dblatex fonts-freefont-ttf \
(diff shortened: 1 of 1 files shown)


ciampix:
Hi @tijuca I agree with you but perhaps, instead of generically say "Ubuntu" or "Debian" we should indicate exactly which version, since Ubuntu 16.04 is a LTS that it will be supported for other three years from now I thing we should leave intact those indications specifying that now they apply for that specific version and for the newer LTS one what it changes ... what do you think about this?

nickoe:
I think this is ok, so I will merge for now.

ciampix:
Good, I will take care of the Ubuntu LTS versions (16.04 and 18.04)

PR #569, opened 2018-05-16
add addendum.zh_CN

oshwcom:


nickoe:
Why do you make pull requests for the 4.0 branch instead of master?

oshwcom:
I look this words in the "readme.adoc":

- Submit your translation pull requests to the stable docs branch 4.0, please.

ciampix:
You are right ... I have to amend it. It meant that you should work on the "stable" documentation, since translating a on going changing text could be frustrating. The thing is that actual "master" is going to be in the near future the next stable, and so it is safe and wise to translate that instead of the stable that will be obsoleted in (hopefully) a short time...

oshwcom:
Thanks.

PR #570, opened 2018-05-16
add chinese translation of CvPcb

CNLHC:
I have translated documentation of CvPcb to Chinese and successfully build it to html .


CNLHC:
@ciampix Thanks，I will be careful next time.

PR #573, opened 2018-05-29
Changed Getting started last copyright year from 2015 to 2018

EeliK:
Certainly someone has edited the document since 2015...


ciampix:
Thanks

PR #585, opened 2018-06-03
Getting Started: Spelling error fix in Bus connections

EeliK:
"on the previous a bus" feels like a spelling error to me, "on the previous bus" is better. I'm not sure if 'a' was meant to refer to the name of the bus, but it's not good anyways.


ciampix:
Thanks!

PR #587, opened 2018-06-04
Try to solve the Non-ASCII language PDF building problem

CNLHC:


CNLHC:
Hello,
I am sure  `a2x` parameters problem mentioned in #580 is a BUG.And I have done some changes on the `cmake` stuff trying to fix it.

PR #590, opened 2018-06-08
Getting Started 5.1. UI string changes and minimal changes to be more…

EeliK:
… in line with version 5.

There were parts which were clearly targeted for the legacy canvas.


ciampix:
Thanks!

PR #593, opened 2018-06-09
Getting Started drillfiles

EeliK:
Modified instructions a bit, drill files were problematic. It should be more logical and complete now.

Changing grid in gerbview was added because while being a basic action there's no main menu or toolbar item for it and may go unnoticed in the context menu.


ciampix:
Very good, thanks!

PR #595, opened 2018-06-10
Getting Started: "component" has been replaced with "symbol" in KiCad.

EeliK:
Another small changes: removed internet link. Added "text editor" to
remove ambiguity.

The link was problematic: the site includes dead links. The search results are also at least partly dead. Unfortunately there aren't good simple alternatives for symbols. There are some commercial alternatives like digikey, ultralibrarian or snapeda with free offerings. Do we want to mention them? We should rather have an up-to-date maintained list in KiCad web page or in forums and give a link to that. I didn't find any for symbols. There's https://github.com/monostable/kicad_footprints for footprints and seems to include some symbol libraries, too.


ciampix:
Thanks, again!

PR #612, opened 2018-07-19
Remove dblatex dependency when building HTML or EPUB only

aimylios:
Cmake always evaluates the PDF_GENERATOR build option, which defaults to "DBLATEX". As a consequence, dblatex and its dependencies need to be installed to build kicad-doc. One can add something like `-DPDF_GENERATOR=none` to the command line to avoid this. Instead, PDF_GENERATOR should only be evaluated if "pdf" (or "PDF", "Pdf", etc.) is included in BUILD_FORMATS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,24 +40,28 @@ add_a2x_option( --no-xmllint )
 add_a2x_option( -f pdf )
 #add_a2x_option( --verbose )
 
-# Add a2x options depending on which PDF generator we're using
-if( ${PDF_GENERATOR} MATCHES DBLATEX )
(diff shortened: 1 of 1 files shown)


aimylios:
Travis failed, although I did not experience any issues when building locally. Might be related to the CMake version. I'll try to find a better solution to my problem.

aimylios:
OK, I updated my proposal to fix the Travis build. This may not be the most elegant solution, but it works.

Please let me know what you think.

nickoe:
Hmm, I have not tested it locally, but it looks resonable and I don't have a better suggestion, so lets merge it. Thanks.

PR #624, opened 2018-09-07
Fix broken link to KiCad upgrade document

### LINKED ISSUE #615: Link broken in migration guide
The first link in the 4->5 migration guide, [kicad_upgrading_from_v4_to_v5.adoc](https://github.com/KiCad/kicad-doc/blob/master/src/kicad/kicad_upgrading_from_v4_to_v5.adoc), is broken when using the "deployed" documentation at [docs.kicad-pcb.org/master/en/kicad.html#_schematic_symbol_libraries](http://docs.kicad-pcb.org/master/en/kicad.html#_schematic_symbol_libraries).

Unfortunately I don't know how to fix this.

The current code works only via github:
```
link:../eeschema/eeschema_symbol_library_table.adoc[lookup table matching
nicknames to library paths]
```

This modified code works only via the deployment:
```
link:eeschema.html#_manage_symbol_libraries[lookup table matching
nicknames to library paths]
```

Is there another way to fix this?

kerel-fs:
Relative links only work in the github-hosted repository.
In the html, pdf and epub documentation outputs this link was broken.

Replacing the relative link with an actual URL fixes this issue.

Fixes #615.

### DIFF
--- a/src/kicad/kicad_upgrading_from_v4_to_v5.adoc
+++ b/src/kicad/kicad_upgrading_from_v4_to_v5.adoc
@@ -21,7 +21,7 @@ would be searched and the first library that held the symbol name
 would be used.
 
 From v5, KiCad symbol names are prefixed with a nickname, and a
-link:../eeschema/eeschema_symbol_library_table.adoc[lookup table matching
+https://github.com/KiCad/kicad-doc/blob/master/src/kicad/kicad_upgrading_from_v4_to_v5.adoc[lookup table matching
 nicknames to library paths] is used to locate the library which holds the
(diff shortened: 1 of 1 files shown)


ciampix:
I do not like URL hardcoded in the docs source code. It seems to me an ugly hack, docs can change source repository or you can be off-line, but I do not see any way to handle it anyway and it works for all the formats so I'm merging it. Thank you for forcing me to make up my mind... sometimes I need a little nudge...

kerel-fs:
Your comment exactly matches my feelings about this change! :)

PR #628, opened 2018-09-26
Update "Making schematic components" link

talos2002:
"2.1. Overview" has a "Making schematic components" link that points to "getting_started_in_kicad.html#make-schematic-components-in-kicad", which doesn't exist.
It is the wrong name and link for "7. Make schematic symbols in KiCad" at "getting_started_in_kicad.html#make-schematic-symbols-in-kicad".

### DIFF
--- a/src/getting_started_in_kicad/getting_started_in_kicad.adoc
+++ b/src/getting_started_in_kicad/getting_started_in_kicad.adoc
@@ -181,7 +181,7 @@ When applicable, the icon is added for convenience.
 image::images/kicad_flowchart.png["KiCad Flowchart"]
 
 For more information about creating a component, read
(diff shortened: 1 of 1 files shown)


ciampix:
Good catch, thanks!

talos2002:
Happy to help!
Should all "schematic components" in the documentation be renamed to "schematic symbols"?

PS: I 'm a total newbie to electronics and KiCad, so don't take me wrong if I 'm mistaken on something.

ciampix:
it depends on context, if I say for example: "In KiCad lingo a schematic component is made of a symbol and one or more associated footprint" it is obviously right ... but probably some old definitions skipped the revision here and there...

PR #632, opened 2018-12-09
Getting Started: images in 4.1 "Using Eeschema"

EeliK:
Images correspond to KiCad master HEAD (dd9a00107 and before, 5.1-dev, in Linux Kubuntu/GTK+ 2.24).
MCU symbol has been updated. Also UI is a bit different, especially the symbol chooser.


ciampix:
Very good, thanks!

PR #637, opened 2018-12-13
Gsg new workflow no netlist for pcbnew

EeliK:
Added information about the new netlist-less workflow in the beginning of the Pcbnew section.


ciampix:
Thanks!

PR #647, opened 2019-01-18
Updated bookmark/section title for Make Schematic Symbols in KiCad

suitengu:
Updated the section jump link and link text in "Using Eeschema", Step 25 to refer to the title "Make Schematic Symbols in KiCad" instead of "Make Schematic Components in KiCad"


ciampix:
Thanks!

PR #650, opened 2019-01-25
Shift-click when adding vias

robert128:
This new add via behavior hasn't been documented anywhere, but is something everyone should appreciate.  I found descriptions of adding vias in many places, so edited those to include the 'Shift' add behavior.

I noticed that the 'Shift' route behavior is described elsewhere in the documentation, but not under 'Creating Tracks' under Manual routing, so added it there.


ciampix:
Good! Many thanks!

PR #659, opened 2019-02-18
Fix：Quotes replace underscores

taotieren:
Fix：Quotes replace underscores


ciampix:
I am going to close this because I already committed a fix to the last translation. I fuzzied a string because it blocked the pdf compilation (I do not know why, it seemed perfectly ok). I am sorry, I hope we will get rid of all this pdf-latex craziness with the migration to asciidoctor...

taotieren:
ok,thanks

PR #665, opened 2019-02-24
add chinese translation for idf_exporter

taotieren:
add chinese translation for idf_exporter


ciampix:
Hi @taotieren, if you would figure out how to exclude commit 9d1fbee from this PR I'll be more that happy to apply it...

PR #639, opened 2018-12-15
Added support for asciidoc semplified Chinese docs

ciampix:

PR #671, opened 2019-02-25
Fix some error for lang-zh.conf

taotieren:
Fix some error for lang-zh.conf

### DIFF
--- a/CMakeSupport/lang-zh.conf
+++ b/CMakeSupport/lang-zh.conf
@@ -15,7 +15,7 @@ ifdef::basebackend-html[]
 
 caution-caption=注意
 important-caption=重要
-note-caption=注释
+note-caption=注
 tip-caption=提示
 warning-caption=警告
 figure-caption=图
@@ -24,12 +24,11 @@ example-caption=例
 toc-title=目录
 appendix-caption=附录
 # Man page NAME section title.
-manname-title=书名
+manname-title=NAME
 
 [footer-text]
-版本 {revnumber}{basebackend-xhtml11?<br />}{basebackend-xhtml11=<br>}
-最后更新 {docdate} {doctime}
-template::[footer-date]
+第 {revnumber} 版{basebackend-xhtml11?<br />}{basebackend-xhtml11=<br>}
+最后更新: {docdate} {doctime}
 
 endif::basebackend-html[]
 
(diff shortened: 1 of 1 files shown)


ciampix:
Please separate the two files...

nickoe:
@ciampix, what do you mean? Each file to two commit?

ciampix:
I mean every commit should not have any conflict. I can approve a commit for file. If one file diff has a problem and the other not, why not commit all those that have no problems and then separately fix the problematic ones? What I would like to avoid is to manually fix every commit that has a problem (if possible). Obviously I can fix it but: 1) if I fix it, the user miss an occasion to learn something and 2) I won't be really sure of the results, since I am afraid I do not know Chinese and I can make mistakes...

taotieren:
Sorry，I want to test the src/idf_exporter/po/zh.po file is not fixed correctly.

taotieren:
[https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=622359;msg=15](url)

taotieren:
There are Chinese Simplified Form (zh_CN) and Chinese Traditional Form (zh_TW) in Chinese. I use Chinese Simplified Form (zh_CN). Chinese translation in kicad-doc is also using Chinese Simplified Form (zh_CN). Can zh.po be updated to zh_CN.po?

taotieren:
Simplified Chinese(zh_CN) and Traditional Chinese(zh_TW) have different vocabulary and usage.

ciampix:
Thanks, I'll rename the file too...

ciampix:
You didn't look at #639? Do you? ;-)

taotieren:
No,I look,Add Single and double quotation marks for lang-zh.conf

ciampix:
Good, thanks. I picked it up that somewhere in the Internet but obviously I was not sure about it's correctness. In a few minutes I'll have it renamed and I'll close all PR related...
I think you can also try to propose the inclusion of that .conf file to the main asciidoc devs so as to have it automatically included in the next asciidoc version...

ciampix:
In your opinion, should we rename all references to zh to zh_CN? I mean, also all the .po files?

taotieren:
Because of China's national conditions, there are two kinds of Chinese, simplified Chinese(zh_CN) and traditional Chinese(zh_TW). At present, all zh.po are simplified Chinese(zh_CN) and can be changed to zh_CN.po.

ciampix:
Yes they can be changed but in your opinion they should? This is a question, I am asking you because I do not know...
 Anyway it is wise to leave it as it is now for the time being (this KiCad version 5.2) since I've seen that cmake scripts do not work with aa_BB type of .po files...

taotieren:
Thanks,I hope to improve this problem in the future.

evanshultz:
@ciampix 
Yes, they should be changed. It's totally ambiguous to not clarify if the translation is Traditional or Simplified Chinese. I took a quick look at https://github.com/KiCad/kicad-doc/blob/master/src/eeschema/po/zh.po, as an example, and the characters are simplified. Our translation should reflect that and leave room for a traditional translation, should one ever be added.

ciampix:
I agree but my knowledge of cmake is near zero. If you change all files from zh to zh_CN cmake will fail ... see  #672 ...

evanshultz:
I understand and I saw that. I'm only answering your question above.

I don't know cmake either.

PR #674, opened 2019-02-27
Update getting_started_in_kicad.adoc

marekr:
Link directly to docs.kicad-pcb.org and reword the sentence above it that was a little awkward.


ciampix:
Next time please apply it after the publication of the current incoming KiCad version and not during the doc & string freeze since this affect all languages as now this string is fuzzied in all language manuals...

nickoe:
@ciampix Yeah, didn't think about that. For this particular change I think we could just unfuzzy the ones that were not fuzzy before. As the text in the translated text are probably not to be changed anyways.

ciampix:
ok I'm doing it.

PR #694, opened 2019-03-10
Fix Chinese typographical errors in simplified parts

taotieren:


taotieren:
Fix translation errors for cvpcb and gerbiew in Simplified Chinese

taotieren:
thanks

PR #705, opened 2019-03-25
Remove libre office meta tags from template creation

### LINKED ISSUE #704: How to make a template section has strange libre office specific lines in html example.
A user just reported on the forum that they are confused about some details of the how to make a template section. Specifically the meta tags found in the html example that i assume are not necessary but are there because the example file was created using libre office.
Any objections on removing these lines?

poeschlr:
Fixes #704 

I also noticed that the html body tag sets the language to french. (But the text is clearly english.)
Is this also ignored by kicad (meaning we can remove this as well) or should we change it to english.

### DIFF
--- a/src/kicad/kicad.adoc
+++ b/src/kicad/kicad.adoc
@@ -500,9 +500,6 @@ Here is a sample *info.html* file:
 <META HTTP-EQUIV="CONTENT-TYPE" CONTENT="text/html;
 charset=windows-1252">
 <TITLE>Raspberry Pi - Expansion Board</TITLE>
-<META NAME="GENERATOR" CONTENT="LibreOffice 3.6 (Windows)">
-<META NAME="CREATED" CONTENT="0;0">
-<META NAME="CHANGED" CONTENT="20121015;19015295">
 </HEAD>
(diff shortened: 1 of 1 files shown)


ciampix:
Thanks!

poeschlr:
What do you think about the language specifier? Should we delete it or change to english?

ciampix:
If you leave it, change it into English. If you think it is worth, open a BR to enable translation of the templates (i.e. handling the language tag...)

poeschlr:
Does kicad even use that attribute? If not then i doubt it would be worth having this in the example even if we want the example html to be translated. (I think the example does not necessarily need translation but that is for others to decide. I have not even looked deep enough into how the documentation stuff works to know how to use translation.)

ciampix:
IHMO we can leave it and add a note that, actually, language tags are not handled (yet?)
Maybe in the future...

nickoe:
If not used, lets just remove it? I mean it is a template afterall.

ciampix:
Yes as I said, _if_ you leave it, then make it correct... but remove any unused tag is generally a good policy

PR #711, opened 2019-04-07
Changed the indication for the Layer setup menu in "Setting up layers"…

svrahul:
…" section

Also added a board thickness information to the description


ciampix:
Thanks!

PR #726, opened 2019-06-26
Correct Copr reference

stevefalco:
The copr reference currently points to mangelajo/kicad, but it should be @kicad/kicad.


nickoe:
Ohh, yeah, definitely. Good catch :)

PR #727, opened 2019-07-09
[pcb_calc] TransLIne references

heitorPB:
I found some of the references for the models used in the Transmission Line Calculator. Some of them are in the source code and some I found in the forums, but I couldn't find all of them. This might be helpful for other people as well.


ciampix:
Very good, thanks!

PR #746, opened 2019-08-03
Update Eeschema_zh_CN_v0.4.1

taotieren:


taotieren:
@ciampix Hi，I have made some updates to the manual. When will it be merged? Or is it a multi-platform compilation test?

ciampix:
In the meantime I merge this... then we should discuss this topic elsewhere, since here is not the right place. The doc dev ml I think is the place...

ciampix:
Thanks!

PR #751, opened 2019-08-06
Fix code block formatting error

follower:
A number of the example code blocks were terminated with one less `-` than needed which meant the formatting for most of the rest of the page was incorrect. This also broke the Footprint Wizard link near the top of the page.

Note: The formatted diff shows extensive changes but there's only a couple of `-` characters added in the raw diff.


ciampix:
Good catch, thanks!

